PR #4, opened 2019-01-18
Player Movement

AustinKellar:
High Level Goal:
Create a system for moving characters via walking, running, jumping and dashing.

Important Files Changed:
PlayerInput.cs: Reads controller / keyboard input using Unity's input manager
PlayerGravity.cs: Applies constant gravity to player object.
PlayerGrounded.cs: Detects when the player is touching the ground
PlayerMovement.cs: Moves the player left and right (walking or running)
PlayerJump.cs: Allows the player to jump
PlayerDash.cs: Allows the player to dash

New Features:

This object is a Unity prefab

**Player Input**
- This script uses Unity's Input manager to capture input from the controller / keyboard
- It houses primarily public properties that are read by PlayerMovement.cs, PlayerJump.cs and PlayerDash.cs
- Player Input Settings (Don't change these)
    - All of the movement axes can be configured inside Unity's input manager (Edit ->Project Settings)
    - When we add more players, we will need to set values inside this component

**Player Gravity**
- This script applies a constant downward velocity on the player object.
- We are not using Unity's built-in rigidbody gravity
- Gravity Settings
    - Gravity: changes the amount of gravity applied to the character

**Player Grounded**
- This script has a public property called IsGrounded which other scripts read from. 
- IsGrounded is true only when the player's Ground Collider object is toching the ground 
- Grounded Settings (Don't change these settings)
  - Ground Collider Layer Mask: Changes what counts as the ground
  - Ground Collider Transform: The transform of the object that collides with the ground (in this case, an invisible sphere attached to the player object)

**Player Movement**
- This script allows the player to move left and right according to user input
- There are 2 speed modes: running and walking
- To walk, hold down the analog stick ~half way to either side 
- To run, hold down the analog stick all the way to the left or right
- Player Movement Settings:
    - Max Running Speed: The limit for how fast the player can run
    - Max Walking Speed: The limit for how fast the player can walk
    - Movement Acceleration: The speed at which the player accelerates to the max speed
    - Movement Drag: How fast the player slows down when they let go of movement controls. This also affects how fast the player can switch directions.

**Player Jump**
- This script allows the player to jump
- Jump Settings:
    - Jump Force: The amount of force applied to the character when you press the jump key
    - Jump Up Gravity Scale: The amount gravity is scaled while traveling up
    - Jump Down Gravity Scale:  The amount gravity is scaled while traveling down
    - Holding Jump Gravity Scalar: The amount gravity is scaled while holding down the jump key
- Note: For all scalars: Small numbers make it more floaty, while large numbers make it more heavy

**Player Dash**
- This script allows the player to dash
- Dash Settings:
    - Reset On Landing: Resets the dash when the player touches the ground (Note: this disables the dash cooldown)
    - Dash Cooldown: The minimum amount of time before the player is allowed to dash again
    - Dash Force: The amount of force applied to the character when dashing
    - Dash Drag: The amount of force applied to slow down the dash
    - Pause Gravity: Pauses gravity during dash
    - Exact Direction: 
        - When true: The player can dash in every possible direction (according to the left analog stick)
        - When false: The player can only dash in 45 degree segments (right, left, up, down, up-right, down-right, up-left, down-left).

How to Test:
- Plug in an XBox controller
- Make sure the controller is set to player 1 (you may need to restart your computer if this isn't the case)
- Without changing settings, make sure you can:
    - Move with the left analog stick
    - Jump with A
    - Dash with B
- Change the settings IN PLAY MODE according to the above descriptions. If you do it in normal mode, your changes will be permanent
    - If you change settings in PlayerInput or PlayerGrounded, movement will likely break (this is intentional)


Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -0,0 +1,186 @@
...
+
+[RequireComponent(typeof(PlayerGravity))]
+[RequireComponent(typeof(PlayerInput))]
+[RequireComponent(typeof(PlayerMovement))]
+[RequireComponent(typeof(PlayerGrounded))]
+public class PlayerDash : MonoBehaviour
+{
+    [SerializeField]
+    [Tooltip("Disables Dash Cooldown")]
+    private bool _resetOnLanding;
```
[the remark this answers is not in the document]
AustinKellar: Our order is actually correct according to the style guide. Our guide has this order:
- public fields
- private fields
-public properties
- private properties

The code that you referenced is a public property, so it should come _after_ any private fields. I believe everything is in the correct order. Given this information, do you still see inconsistencies?

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -0,0 +1,186 @@
...
+            else
+            {
+                ApplyVelocityInLimitedAngles();
+            }
+        }
+    }
+
+    private void ApplyVelocityInLimitedAngles()
+    {
+        if (_playerInput.HoldingRight && _playerInput.HoldingUp)
```
IanGRap: This is a lot of if statements. What if you set x as a variable based of if we are moving left or right and set y as a variable based off up and down. Should be way less lines.

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -0,0 +1,186 @@
...
+    {
+        Dash();
+    }
+
+    private void FixedUpdate()
+    {
+        if (IsDashing)
+        {
+            AddDrag();
+        }
```
IanGRap: Minor, but maybe it is worth putting this if check in AddDrag() to minimize code in FixedUpdate

Assets/Scripts/Player/PlayerGrounded.cs
```diff
@@ -0,0 +1,21 @@
...
+    private LayerMask _groundColliderLayerMask;
+
+    private bool _isGrounded;
+
+    public bool IsGrounded { get { return _isGrounded; } }
+
+    private void Update()
+    {
+        _isGrounded = Physics.OverlapSphere(_groundColliderTransform.position, 0.5f, _groundColliderLayerMask.value).Length > 0;
+    }
```
IanGRap: I believe fixed update gets called more than update and since there is no input checking here I would recommend moving this code into FixedUpdate().

Assets/Scripts/Player/PlayerDash.cs
AustinKellar: I'm not sure I understand. Could you lay out some example code?

Assets/Scripts/Player/PlayerGrounded.cs
AustinKellar: That was our thought as well. We originally had this in FixedUpdate but occasionally, the player would not jump when the key was pressed.

Assets/Scripts/Player/PlayerDash.cs
AustinKellar: Can do, we debated on doing that.

Assets/Scripts/Player/PlayerDash.cs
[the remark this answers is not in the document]
egomish: Ah, I see, I misunderstood the C# syntax. Thank you for the clarification.

PR #5, opened 2019-01-22
Block pushing

IanGRap:
High Level Goal:
Create a system of blocks for falling, pushing and collision detection.

Credit:
Ian Rapoport and James Coulter Petnic

Important Files Changed:
BlockController.cs: The main course that manages block movement for falling and pushing as well as check collisions.
BlockManager.cs: Has designer accessible variables for tweaking values like block push speed and gravity acceleration for all blocks in the scene.
BlockTester.cs: Temporary code for testing blocking pushing using keyboard input without character interactions.

New Features:

A Block prefab.

BlockController.cs:

This script controls the main functionality for blocks. In the air, they accelerate toward the ground.If they get pushed left or right they will move 1 unit in that direction while still falling. If they get pushed up or down they will move 1 unit in that direction, ignoring gravity while doing so. If there is a block in the way when being pushed left or right, the push will be cancelled. If there is a block or the floor in the way when being pushed up or down the block will continue until it collides. Pushing multiple blocks, having blocks break when they collide with a wall or having the player be able to push blocks has not yet been implemented.

BlockManager.cs: 

This script contains properties that are accessed by every block in the scene. They include the maximum magnitude of gravity, the acceleration of gravity, the the speed at which blocks get pushed and the side length of the blocks in the scene. These values can each be edited in the Unity editor window. There must be a BlockManager in the scene or else the blocks will print an error and use default values for the previously listed variables.

BlockTester.cs:

This is a temporary script used to test block pushing since player interaction has not yet been implemented. It has a public reference to 1 block in the scene and can call push in any direction on that block.

How to Test:

Load either BlockTesting scene. Press play and use W, A, S, and D to move the pink block on the screen up, left, down and right while it falls. The block should obey the rules outlined in the BlockController.cs description. You can adjust the values of the block by changing the values on the Block Manger gameobject. These updates should affect every block in the scene.


AustinKellar:
Reviewers: Austin and Alfredo

Overall:
The feature works and I think it will get the job done in the short term but we think it needs some more iteration. We think that the code could be separated out into different scripts to make it more readable. For example, Pushing, Ground Detection and Rigidbody Constraints could be their own files. This might require making public gets on key variables such as IsGrounded, IsPushed, PushDirection, etc. We think this code is a bit overly complex. We have a feeling that it could be simplified with more iteration.

Unexpected Behaviors While Testing:
- Blocks being pushed up move slightly more than 1 unit (if 1 unit is the size of a block)
- While falling and being pushed up, if a block runs into the one above it, the above block moves slightly
- Changing serialized values in play mode has no effect. Is this intentional?
    - Personally, I think it makes it harder to test
- When Side Length is set to 2, after you hit the ground, movement sometimes (maybe always) gets locked

IanGRap:
Austin these sound like good suggestions. We'll discuss how to go about implementing them in the next programming meeting.

PR #8, opened 2019-01-27
Squish Detection

AustinKellar:
Contributors: Austin

Goal:
Implement player death when they get squished by a block

Important Files Changed:
- SquishDetector.cs

Changes:
- Added a child object to the player prefab called SqishDetector
    - It is just a trigger sphere collider with a script attached 
    - SquishDetector.cs: Keeps track of all blocks currently overlapping with the player object and destroys 
      the player if they remain overlapped with a block for too long (~0.1 seconds is the sweetspot, I think)

Customizable Fields within Unity:
- Squish Sensitivity: This determines the scale of the Squish Detection collider. When the number is larger, it will be easier to get squished by the corner of a block. When the number is low, if the corner of a block collides with the player, they will likely be pushed out of the way instead of killed. Crank up the sensitivity to make the game more punishing
- Squished Buffer Time: This value probably won't need to be changed. While I was testing, I noticed that sometimes, Unity's physics would cause the player to dash into the blocks for about a frame. This was causing unintended squishing, so I added a buffer time. Right now, the block must be overlapping with the squish detector for 0.1 seconds in order for a squish to occur. If you want to see why this field is needed, try setting it to 0 and dashing into a few blocks (you will likely die).

How to test:
I created 2 scenes inside Assets/Scenes/Squish Detection. Try getting squished in both:

1) Control the Player: Here, you can control the player as blocks are falling from above. You can get squished only from above in this scene.
2) Control the Block: Here, there are 4 player objects, all with keyboard input toggled off. There is also 1 block with keyboard input toggled on. In this scene, you can test squishing the player from above, blow, left and right. 

After testing basic squishing, modify Squish Sensitivity and try to get squished by just the corner of blocks. I think a decent sensitivity is 0.75 but it should be customizable.


AustinKellar:
> Looks good to me. I'm curious why we need to make a new sphere collier for squishing as oppose to using the one that already exists for collision.

I originally used the same collider but it detects a squish when the player is standing on top of a block. Having a separate collider for squish detection also allows for sensitivity customization. This would not be possible (or at least not easy) if everything used the same collider.

AlfredoCuevas:
It seems to work as intended. 

Notes: Things I noticed but probably aren't issues
1. I noticed that when two players are standing next to each other and a block pushes them into a wall they seem to merge into one 1x1 space. But if you move the block back and do it a second time then one of the players dies where before they merged into the 1x1 space. 
2. Sometimes if a block gets pushed into the player in rapid succession they will get squished without being stuck between two objects. This probably won't matter because there will be a cooldown on the dash. Still thought I would mention it.

AustinKellar:
These sound like bugs with player movement / block pushing to me. I'll go ahead and merge this. We just need to make sure we track these bugs

PR #9, opened 2019-01-28
Push detection

AustinKellar:
Contributors: Austin

Goal:
Made block pushing w/ player dash more accurate

What I Changed:
- I played with the block prefab, altering the dimensions of the player colliders so that they are longer and narrower. This ensures that the player cannot collide with more than one side of the block at a time.

How to test:
- Pull development (not this branch) and go to Assets/Scenes/Block PR/Multiple Blocks With Player
- While the blocks are still falling, try pushing the blocks to the side. 
    - Often, the blocks will be pushed in the wrong direction. I.E. sometimes when you collide with the right side of the block, it will go up instead of to the left. 
- After playing with the old feature, pull this branch and test in the same scene. The block pushing should be more accurate. I.E. when you collide with a block on the right side, it should go left ~every time.


IanGRap:
Feels good!

PR #11, opened 2019-01-30
Squish detection bugfix

AustinKellar:
Goal:
Fix the bug where the player would sometimes die when moving quickly, even if they weren't squished

Important Files Changed:
- The Player prefab:
    - Instead of 1 squish collider, there are 4. (Top, Bottom, Right and Left)
- SquishDetector.cs

What Changed:
- Squish detection needed to be a lot more complex to fix this bug. I added in the following checks before death:
    - (Top + Bottom) colliders or (Right + Left) colliders must be activated in tandem in order to trigger a squish
    - The player must be colliding with at least 2 _unique_ objects in order to trigger a squish
    - The player will not be squished if they are between 2 objects that are less than the Min Squish Distance apart (1.5 units at this time)
        - This prevents the scenario of the player dying when Unity physics clips them directly between 2 blocks that are clumped together
    - The player cannot be in the middle of a dash in order to trigger a squish

How to Test:
- Open Assets/Scenes/demos/sprint 1
- Dash into blocks from all angles, trying to cause a player death without actually being squished between 2 objects
- Make sure that you do still get squished when you _are_ supposed to!

Note: 
- I played the scene all the way through about 15 times. Out of those tests, I got only 1 questionable death. This squish detection is definitely more accurate but still (I think) not perfect. However, I'm fresh out of ideas of how to make it more perfect. Perhaps someone else will have some ideas. I'm definitely open to suggestions.


Assets/Scripts/Player/SquishDetector.cs
```diff
@@ -3,57 +3,65 @@
...
-        }
+        _playerDash = GetComponent<PlayerDash>();
     }
 
     private void Update()
     {
-        _overlappingBlocks = _overlappingBlocks.Where(obj => obj != null).ToList();
-
-        if (_overlappingBlocks.Count > 0)
+        if (!_playerDash.IsDashing)
```
IanGRap: Personally, I would change this to an early return so the entirety of our function isn't in an if statement

Assets/Scripts/Player/SquishDetector.cs
```diff
@@ -3,57 +3,65 @@
...
+                ).ToList();
+
+                GameObject left = _leftCollider.OverlappingObjects.First();
+                GameObject right = _rightCollider.OverlappingObjects.First();
+                float squishDistance = Vector3.Distance(right.transform.position, left.transform.position);
+                if (pushedBlocks.Count> 0 && left != right && squishDistance > _minSquishDistance)
+                {
+                    Destroy(gameObject);
+                }
+            }
```
IanGRap: Update is looking very cluttered. Can we refactor this into a separate function that update calls?

Assets/Scripts/Player/SquishDetector.cs
```diff
@@ -3,57 +3,65 @@
...
+                List<BlockPusher> pushedBlocks = (
+                    from b in blockPushers
+                    where b.IsPushed && (b.PushDirection == Vector3.left || b.PushDirection == Vector3.right)
+                    select b
+                ).ToList();
+
+                GameObject left = _leftCollider.OverlappingObjects.First();
+                GameObject right = _rightCollider.OverlappingObjects.First();
+                float squishDistance = Vector3.Distance(right.transform.position, left.transform.position);
+                if (pushedBlocks.Count> 0 && left != right && squishDistance > _minSquishDistance)
```
AlfredoCuevas: I'm curious why you need the check (pushedBlocks.Count > 0). Is there some kind of bug that gets introduced when you don't check that the block is being pushed?

AustinKellar: Yes, if the blocks get a little offset and the player squeezes between them, they could get squished, even if no blocks have been pushed.

Assets/Scripts/Player/SquishDetector.cs
AustinKellar: Will do

PR #13, opened 2019-02-01
Nonstick Player!

AustinKellar:
Goal: 
Fix a bug where the player sticks to walls if they are in the air and holding horizontal movement towards that direction

Important Files Changed:
- PlayerMovement.cs: 
    - Cancel walking if 
        - The player is touching an object to the left trying to move left
        - The player is touching an object to the right and trying to move right 
- PlayerAgainstWall.cs
    - Uses raycasts to determine if the player is against a wall/block on either the left or right side

How to test:
- Open the sprint1 demo scene
- Jump into walls while holding movement controls and try to stick to the wall
    - You should be unable to stick

Note: 
-I added a new Layer to unity called Wall. In order for the nonstick to work, all walls need to be assigned the "Wall" layer. Also, blocks must be assigned the "Block" layer. I only applied the wall layer on the sprint1 demo scene.


JamesCoulter:
I wasn't sticking to walls. I wasn't able to find any issues or other bugs, so I'd say it's all good.

PR #21, opened 2019-02-05
Better camera

AlfredoCuevas:
Contributors: Ian, Alfredo
This Brackeys video helped: https://youtu.be/aLpixrPvlB8

Goal: 
Create an improved camera that tracks the action of the game

Important Files:
CameraManager.cs: Calculates the camera parameters based on players and editor assigned values. CameraManager is an instance that can be called from anywhere.

SquishDetector.cs : Added one line to tell the camera to stop tracking this player after being squished

How To Test: 
* Load up BetterCamera scene.
* Play through the game. The camera should update its position and change its field of view based on how far apart the players are
* The players should always be visible
* The camera adjustment should not be jarring or nauseating

Notes: 
* Min/Max FOV: The range of field of view values for the camera
* Map Width/Height: should be approximately the size of the area of play
* Min Camera Height: Lowest height the camera can go.
* These values should only need to change based on the size of the map. 

Public Functions: Can be called through the Instance of the CameraManager
*  AddTrackedTransform(): Allows you to add an object to be tracked by the camera
* RemoveTrackedTransform(): Allows you to remove an object that is tracked by the camera
* AddTemporaryTransform(): Allows you to add a transform to be tracked for given amount of seconds
 **The add functions are not currently being used but may be useful later**


Assets/Scripts/CameraManager.cs
```diff
@@ -0,0 +1,121 @@
...
+    public void RemoveTrackedTransform(Transform removedTransform)
+    {
+        _trackedTransforms.Remove(removedTransform);
+    }
+
+    private void SetBounds()
+    {
+        _bounds = new Bounds(_trackedTransforms[0].position, Vector3.zero);
+
+        for (int i = 0; i < _trackedTransforms.Count; i++)
```
AustinKellar: This is nitpicky, but can we use a foreach here?

Assets/Scripts/CameraManager.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        _trackedTransforms = new List<Transform>();
+    }
+
+    private void Start()
+    {
+        _trackedTransforms = _trackedTransforms.Concat(GameObject.FindGameObjectsWithTag("Player").Select(obj => obj.transform)).ToList();
+    }
+
+    private void LateUpdate()
```
AustinKellar: Just curious, why LateUpdate?

Assets/Scripts/CameraManager.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            return;
+        }
+
+        SetBounds();
+        SetPosition();
+        Zoom();
+    }
+
+    public void AddTemporaryTransform(Transform newTransform, float seconds)
```
AustinKellar: I noticed that this isn't called anywhere. Why do we have it? Is it looking ahead towards when we have items?

Assets/Scripts/CameraManager.cs
AlfredoCuevas: You let all of the player positions update first in the regular Update() statement. The camera comes right after in the LateUpdate() so you can be sure everything else is positioned where it needs to be.

Assets/Scripts/CameraManager.cs
IanGRap: Exactly. We didn't see a use for it now but it could be cool later.

PR #27, opened 2019-02-07
Player Selection

AustinKellar:
Contributors: Alfredo Cuevas, Austin Kellar

Goal: Create a system for selecting the number of players and starting a match with the correct number of players, all hooked up to the correct controller

Important Files Changed:
- NEW SCENE: Player Selection/Player Selection Screen (contains player select UI and logic for mapping players to controllers)
- NEW SCENE: Player Selection/player selection game test (a copy of sprint 2 demo which has four players in it)
- SceneController.cs (a static object which allows you to choose which scene gets started when players 
press Start in the player selection screen)
- ActivePlayers.cs (a static class that can be accessed globally which contains information about which players are signed in to play in the next match)
- UIController.cs (Controls the UI text in the player selection scene)
- ControllerManager.cs (Assigns controllers to players)
- Controller.cs (Holds information about controllers)
- Player.cs (Holds information about players in relation to what controller they are assigned)
- PlayerManager.cs (Adds and removes players from ActivePlayers.cs and updates the player select UI)
- PlayerInput.cs (On awake, if there is no player matching the player's assigned PlayerNumber inside ActivePlayers.cs, delete that player)
- ControllerManager.cs (Make sure that the camera is not trying to track null objects)

How to Test:
- Load the scene Player Selection Screen
- Connect 2-4 players and press Start
- Ensure that the player selection game test scene is loaded with the correct number of players
- Ensure that for each controller that joined the match, that controller move exactly 1 of the players in the match

Notes:
- You have to choose what scene will start when you press Start (inside the Scene Controller object)
    - The scene that you choose needs to exist in the Build Settings or you will get a console error
- We got rid of the input fields inside PlayerInput.cs and replaced them with a single field called PlayerNumber. All previously existing scenes that have Players in them WILL NOW BE BROKEN. You will have to assign each player in the scene a player number from 1 to 4 in order for controller input to be recognized.
- There is still the bug where if you disconnect controllers and reconnect them, unity will think that they are a different controller besides 1, 2, 3, 4. We have a potential fix in mind for this but it will take a lot of time to implement so we think it should be its own PR


Assets/Scripts/Player Selection/PlayerManager.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public void ReorderPlayers()
+    {
+        _players.ForEach(p => MovePlayerToFirstOpenSlot(p));
+    }
+
+    private void MovePlayerToFirstOpenSlot(Player player)
+    {
+        if (player.HasController)
```
IanGRap: early return?

Assets/Scripts/Player Selection/PlayerManager.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+public class PlayerManager : MonoBehaviour
```
IanGRap: Should there be a check that there is only 1 of these in a scene?

Assets/Scripts/Player Selection/ControllerManager.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+
+public class ControllerManager : MonoBehaviour
```
IanGRap: Should there be a check that there is only one of these in a scene

Assets/Scripts/Player Selection/ControllerManager.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    private void Update()
+    {
+        _controllers.Where(c => !c.IsAssigned && c.JoinButtonDown()).ToList().ForEach(c => c.AssignController());
+        _controllers.Where(c => c.IsAssigned && c.LeaveButtonDown()).ToList().ForEach(c => c.RemoveController());
+        
+        if (_controllers.Where(c => c.IsAssigned).ToList().Count >= 2)
+        {
+            if (_controllers.FirstOrDefault(c => c.IsAssigned && c.StartButtonDown()))
```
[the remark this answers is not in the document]
AlfredoCuevas: No. It means any **active** player you presses the start button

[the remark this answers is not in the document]
IanGRap: It wasn't working with my second controller but I was also using a ps4 one so I will need to test in lab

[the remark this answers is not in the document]
AlfredoCuevas: Yea we only set up xbox controllers for now. It could be that playstation start button is mapped to a different button in the Input menu

Assets/Scripts/ActivePlayers.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+public static class ActivePlayers
```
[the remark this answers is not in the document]
AlfredoCuevas: I don't think it needs an instance if it is a static class. It works in a similar way as if we had a static instance of a class except this doesn't need a Unity Object in the scene

Assets/Scripts/Player Selection/ControllerManager.cs
```diff
@@ -0,0 +1,36 @@
...
+        _controllers = GetComponentsInChildren<Controller>().ToList();
+
+        int index = 1;
+        _controllers.ForEach(c => c.SetIndex(index++));
+    }
+
+    private void Update()
+    {
+        _controllers.Where(c => !c.IsAssigned && c.JoinButtonDown()).ToList().ForEach(c => c.AssignController());
+        _controllers.Where(c => c.IsAssigned && c.LeaveButtonDown()).ToList().ForEach(c => c.RemoveController());
```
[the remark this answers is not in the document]
AlfredoCuevas: Gotta give austin credit for this one

Assets/Scripts/Player Selection/PlayerManager.cs
AlfredoCuevas: I'm not sure what you are asking. What this function is doing is reorganizing the player list. So if Player 1 is assigned, Player 2 is unassigned, and Player 3 is assigned then it will make Player 3 become Player 2.

IanGRap: Its a style thing. The entire function is in an if statement so I usually have if not return, otherwise do the rest of the function.

AlfredoCuevas: Is that more clear to ask if they don't have a controller than if they do?

PR #28, opened 2019-02-08
Block clearer

IanGRap:
Contributors: Ian

Goal: Create a prefab that clears blocks once they reach the bottom of the screen.

Important Files: BlockClearer.cs. Contains the code for block clearing. Map width is the number of columns in the map.  Clear on Row Filled decides if the bottom lane will be cleared when the entire row has filled up, otherwise clear all blocks after a set number of time. Clear Timer is the interval of time that the rows are cleared on.

How to Test: Load up the Block Clearer scene. Find the block clearer prefab. Set it to clear on filled and the bottom row should disappear when it fills with blocks. Turn off clear on filled and all the blocks on the bottom row should clear on the timer you set.

Notes: I think both of these approaches can be expanded upon, but this is a solid placeholder for playtesting.


Assets/BlockClearer.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+        foreach (GameObject block in _overlapBlocks)
+        {
+            Destroy(block);
+        }
+        _overlapBlocks.Clear();
+        AudioManager.Instance.PlayOneShot("Block Collision Rock");
+    }
+
+    private void ClearRowOnTimer()
```
AlfredoCuevas: Same with this it feels a little cramped to me.

AlfredoCuevas: Lol, I messed up. It didn't keep my other comments. Essentially I just mentioned that the last two private functions felt a little cramped and I thought they would benefit from some spaces between certain {blocks} of code.

IanGRap: haha for sure I'll update them

PR #41, opened 2019-02-13
Block Acceleration

IanGRap:
Contributors: Coulter, Ian

Goal: Add in functionality for blocks to accelerate with gravity instead of fall at a constant speed.

Important Files:

- BlockMovement.cs: When gravity is calculated it is accelerated overt time towards a terminal velocity. Also added public functions to get and set the speed of gravity.

- BlockAdjacenyQueries.cs: Added new check to see if below the block is obstructed by the ground, a stationary block or a block being pushed.

- BlockPusher.cs: Adjusted the buffer/timer values to mitigate when the block gets pushed but doesn't move.

- GameManager.cs: Changed when music is activated to avoid awake vs start null problems.

- Sprint 2.scene: Removed the box collider from the roof to stop problems where blocks are smooshed when they are spawned.

How to Test: Load up the Sprint 2 scene and adjust the BlockAcceleration and TerminalVelocity values on the Block prefab. The terminal velocity is the max speed the block will fall at. The acceleration is how much the speed will increase each second. Try experimenting with the values. If you want to have a constant velocity for gravity, set the acceleration very high (maybe 30).

Notes: 

- I changed some stuff in the sprint 2 demo scene even though I said I wouldn't because I am CRAZY.

- Sometimes the blocks wouldn't move when you push them. This has hopefully been fixed but if it happens again we will need to tweak the push values again.


Assets/Scripts/Blocks/BlockAdjacencyQueries.cs
```diff
@@ -47,13 +48,34 @@ public bool AdjacentBlockIsPushed(Vector3 direction)
...
+        }
+        else
         {
             return false;
         }
     }
 
+    public bool FallingPathBlocked()
+    {
+        foreach(GameObject obj in GetAdjacentObjects(Vector3.down))
```
AustinKellar: I think this would be cleaner using a linq query but no big deal.

Assets/Scripts/Blocks/BlockMovement.cs
```diff
@@ -26,24 +29,27 @@ private void Awake()
...
         ApplyGravity();
     }
 
     private void Update()
     {
-
         ClampToDestination();
     }
 
+    public float GetGravitSpeed()
```
AustinKellar: This function name is misspelled

PR #44, opened 2019-02-14
Bugfix Controllers

AustinKellar:
Goal: Make the controllers more consistent by using an asset from the Asset Store instead of using Unity's prebuilt input system

Important Files Changed:
- PlayerInput.cs (Converted inputs to new format)
-MatchManager.cs (Converted inputs to new format)
- Controller.cs (Converted inputs to new format)

How to Test:
- Load up the sprint 2 demo scene and make sure controls behave the same way they used to
- Do the same with the sprint 2 player select scene
- Unplug and replug controllers a bunch of times. Make sure they find their way back to their original assignments


AlfredoCuevas:
If possible i'd like for someone to test with different controllers like ps4.

PR #45, opened 2019-02-15
Basic Block Particle Effect

JamesCoulter:
Contributors: James Coulter, Ian

Goal: Implement Block break and player death particle effects to improve feedback from game and enhance player experience.

Files Changed: New prefabs for particle effects created, added to players and blocks in editor.

How to Test: Run the Sprint 2 scene in demo build folder. Push block into ground or against wall to break it, you should see small bricks and dirt trails. Let players be squished, and you should see particles emanate outward from their position that match their color.

Notes: There is a merge conflict that needs resolving.


JamesCoulter:
@AustinKellar How much smaller are you thinking?

Assets/Scripts/Environment Actions/BlockClearer.cs
```diff
@@ -87,6 +90,7 @@ private void ClearRowOnFilled()
             Destroy(block);
         }
 
+        Instantiate(_clearEffect, transform.position, Quaternion.identity);
```
AlfredoCuevas: I get an error because it says that there is no particle effect for a block clear. Either make sure the particle effect is properly referenced or take this part out.

AustinKellar:
Maybe 5, 6 times smaller? Small enough that they are in the background when they explode. Right now, it really grabs attention

Assets/Scripts/Environment Actions/BlockClearer.cs
JamesCoulter: The error shouldn't show up anymore

Issue #47, opened 2019-02-18
controllers not working for standalone build


IanGRap:
Description: While controllers appear to be working in the Unity Editor, they do not appear to be working upon creating standalone builds. Upon loading up the character select screen, controllers only work some of the time and generally will not have their inputs read. Unplugging and re plugging controllers repeatedly seems to help but isn't ideal.

Reproduce: Build a standalone version for windows, begin play and see if your controllers connect.

IanGRap:
fucking closed!

PR #49, opened 2019-02-18
Bugfix #47 (controllers in standalone builds)

AustinKellar:
Goal: Fix Issue #47 

What I changed:
- In ControllerManager.cs on awake, I added a line of code which reinitializes all controllers

Note:
- I don't know if this will help anything. It is more of an experiment


AustinKellar:
Yeah, I'm not convinced this actually helps anything either.

IanGRap:
I want to cry. Lets talk to Super Sumo tomorrow at the play test and see if they ran into anything similar

IanGRap:
Maybe we should post on like the Unity forum

PR #51, opened 2019-02-19
Garbage compactor

IanGRap:
Contributors: Ian

Goal: Creator a trash compactor mechanic that decreases the size of the play are over time and makes it more difficult for players in the late game to survive.

Important Files:

- WallManager.cs: Keeps track of the walls in the scene and tells them when to start shaking and move inward based on a timer;

- Wall.cs: Plays a shaking animation to alert players when it is about to compact and moves over one column unit when compacting.

- MatchManager.cs: No longer tracks the walls in the scene.

- BlockSpawner.cs: Gets References of block locations from WallManager instead of MatchManager:

- BlockClearer: Checks for if a row is full are not dependent on the arena size dictated by the WallManager.

- BlockDamager: Deals with functionality regarding block damage. Right now it only destroys the block.

-BlockPusher: Calls BlockDamager when pushed into a wall.

-Player.prefab: Froze y rotation on rigidbody. Without this the player spins whenever they touch the walls.

- SquishDetector.cs: Squished from the side now ignores whether or not a colliding block has been pushed in the wall the player is touching is trash compacting.

How to test: Load up the trash compactor scene. Look for the wall manager in the scene. 

- Minimum Area Width: The minimum number of columns the arena can compact to.
- Compactor Timer: The number of timer between each compacting of the walls. If you set this too low the compactor won't be able to finish before it is activated and will error out.

The walls should move in periodically during play and destroy any blocks next to them as the do so. Blocks should always spawn within the walls. If the wall compacts on a player they should be squished.

Notes:

- The wall art is placeholder.
- This one is big so I wouldn't be surprised if there are some hidden bugs.


Assets/WallManager.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class WallManager : MonoBehaviour
+{
+    [SerializeField]
+    Wall _rightWall;
```
AustinKellar: Can we make these private?

Assets/Wall.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Wall : MonoBehaviour
+{
+    [SerializeField]
+    float _wallHeight = 20f;
```
AustinKellar: Can we make these private?

Assets/WallManager.cs
IanGRap: will do

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -64,9 +53,7 @@ public bool Push(Vector3 direction)
         {
             if (_breakOnEnvironmentCollision)
             {
-                AudioManager.Instance.Play("Block Break Rock");
```
AlfredoCuevas: I'm curious why you decided to make a whole new class for this instead of just creating a private function

IanGRap: I was assuming we could expand on it for block health

PR #61, opened 2019-02-20
Making sprint 3 scene

IanGRap:
Contributors: Ian

Goal: Set up a sprint 3 scene with our changes thus far.

Important Files:
- Block.prefab: Set all the colors back to pink to avoid confusing players. (Addresses Issue 57)
- Wall.prefab: Set the box collider dimensions on the prefab instead of in code. It was having problems when the collider overlapped with the floor.
-Wall.cs: Removed references of wall height in code. Fixed bug where walls wouldn't move to right positions.
-MatchManger.cs Addressed bug where music would repeat.

How to test: Load up the Sprint 3 character select and sprint 3 scenes.


IanGRap:
Coulter the player select should load the correct scene now

PR #62, opened 2019-02-21
Added block health and color change

JamesCoulter:
Contributor: Coulter

Goal: Add a health value to blocks so that they break after being pushed a certain number of times. Also create a visual indicator for health being lowered.

Files Changed:
- BlockDamager.cs - Added variables and method to track health, modify color.
- BlockPusher.cs - call BlockDamager.cs methods on push to decrement health.

How To Test: Start any scene with the blocks and push them. After a certain number of pushes (the value of which is exposed in the BlockDamager script), they should break, and each push should cause their color to change noticeably.


Assets/Scripts/Blocks/BlockDamager.cs
```diff
@@ -4,13 +4,42 @@
...
 public class BlockDamager : MonoBehaviour
 {
+    [SerializeField]
+    private int _pushLimit = 3;
     [SerializeField]
     private GameObject _breakEffect;
-    
+    [SerializeField]
+    private float _percent = 0.1f;
+    private int _pushCount = 1;
```
IanGRap: shouldn't the push count start at 0 if you haven't been pushed yet?

JamesCoulter: Fixed, not sure what my initial thought process was.

PR #73, opened 2019-02-27
Random starting blocks

AustinKellar:
Contributors: Coulter, Austin

Goal:
- Implement a feature where the configuration of blocks and players changes every time you start a new match

Important Files Changed:
- MatchManager.cs 
    -Calls SceneCreator function to create the starting configuration / Calls PlayerColorController function to assign colors to the players
- Materials.cs 
    - static class that stores all potential colors of players
- PlayerColorController.cs
   - Assigns unique color to each player at the start of the match
- SceneCreator.cs
    - Chooses a configuration prefab, then instantiates player and block objects according to their locations in the configuration prefab
- SquishDetector.cs
    - Set the color of the particles according to the color of the player's material

How to Test:
- Load up the 2017 scene and make sure that you get different starting configurations each time you restart the match
    - This should work when you press Restart from the pause menu, Press Start to Play Again, and start the match from PlayerSelect
- Make sure that the functionality also works in a standalone build


Assets/Scripts/Managers/SceneCreator.cs
```diff
@@ -0,0 +1,49 @@
...
+        structure.transform.position = Vector3.zero;
+
+        List<Transform> playerTransforms = structure.GetComponentsInChildren<PlayerInput>().Select(p => p.transform).ToList();
+        List<Transform> blockTransforms = structure.GetComponentsInChildren<BlockPusher>().Select(b => b.transform).ToList();
+
+        foreach(Transform player in playerTransforms)
+        {
+            player.gameObject.SetActive(false);
+            GameObject obj = Instantiate(_playerPrefab, player);
+            obj.transform.parent = structure;
```
IanGRap: couldn't we just get the position of all the players then set the new players to that position. Since the parent object is already at 0 the local positions are also the world positions

AustinKellar: That's what we are doing here. I'm not sure I understand what you mean

Issue #15, opened 2019-02-01
horizontal dash is slower and more powerful than vertical or diagonal dash


AustinKellar:
Description:
- With low dash force and low dash drag, a horizontal dash will move the player farther than any other direction.

Reproduce case:
- Go to a scene that has a player object in it
- Set the player's Dash Force to 20 and Dash Drag to 0.05
- Dash in the Up-Right direction. Notice how far the dash takes the player and how long it takes
- Next, dash in the Right direction. The dash takes longer to complete and the distance traveled is greater.

PR #76, opened 2019-02-28
Bug fix: Issue #15 (Dashing diagonally not working right)

AlfredoCuevas:
Contributors: Alfredo

Goals: Fix bug where dashing does not work the same vertically vs horizontally

Files Changed: 
PlayerJump.cs: Added a check to IsTravelingUp/ IsTravelingDown so that they don't apply during dashing
PlayerDash.cs:
- Added a new serialized field that allows designers to check whether they want the dash direction to be normalized when applying force
- Added a check when dashing to make sure that the player was holding the left stick in a direction when pressing the dash button. This way it doesn't show the dash animation/sound when standing still

How to Test: 
* Load up the 2017 scene. Try dashing up and try dashing right. They should feel like they are moving about the same distance. 
* Dash diagonally. It should feel like the amount of vertical and horizontal force are the same. 
* Try disabling **Normalize Diagonal Dash** in the PlayerDash.cs prefab. This should make it so that diagonal dashing travels the same amount of horizontal / vertical distance as when you you aren't dashing diagonally. This means that the player is traveling farther overall. 
* * Note: Normalizing diagonal dashing was the original design choice but I made it an option to address the description of the bug. 
* Lastly try dashing in place. This should no longer show the dashing animation / sound. 

Also here is the reproduce case described in the bug if you want to test this as well: 
Reproduce case:

* Go to a scene that has a player object in it
* Set the player's Dash Force to 20 and Dash Drag to 0.05
* Dash in the Up-Right direction. Notice how far the dash takes the player and how long it takes
* Next, dash in the Right direction. The dash takes longer to complete and the distance traveled is greater.


Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -107,7 +118,7 @@ private void Dash()
             _groundedSinceLastDash = true;
         }
 
-        if (_playerInput.PressDash && !IsDashing && (!_resetOnLanding && DashOffCooldown || (_resetOnLanding && _groundedSinceLastDash)))
+        if (_playerInput.PressDash && IsHoldingADirection && !IsDashing && (!_resetOnLanding && DashOffCooldown || (_resetOnLanding && _groundedSinceLastDash)))
```
AustinKellar: Just out of curiosity, why do we need the check for IsHoldingADirection?

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -158,7 +169,15 @@ private void ApplyVelocityInLimitedAngles()
 
     private void ApplyInitialDashVelocity(Vector3 direction)
     {
-        _rigidbody.velocity = Vector3.Normalize(direction) * _dashForce;
+        if (_normalizeDiagonalDash)
+        {
+            _rigidbody.velocity = Vector3.Normalize(direction) * _dashForce;
```
AustinKellar: Were we not already normalizing?

Assets/Scripts/Player/PlayerDash.cs
AlfredoCuevas: I was having issues testing because we had a small bug where if you click dash but you aren't holding a direction then you dash but don't move. So I decided to fix to get it out of the way. It isn't actually necessary for fixing the vertical dashing issue but it made it easier.

AustinKellar: Oh, cool. Makes sense.

Assets/Scripts/Player/PlayerDash.cs
AlfredoCuevas: we were but the bug was complaining about horizontal dashing traveling further than diagonal dashing so I added a bool to let people choose.

AlfredoCuevas: By default it still chooses the same functionality that we already had.

AustinKellar: Ah, cool. Nice to have extra customizability

AustinKellar: My goodness, you totally said that in the description but I missed it.

PR #90, opened 2019-03-03
Added dash trail to player

IanGRap:
Contributors: Ian

Goal: Add a dash trail vfx for when the player dashes to increase JUICE!

Important Files: 
- PlayerColorController.cs: Sets the material of the trail to be the material of the character.
- PlayerDash.cs: Activates and deactivates the dash trail when the player starts and finished dashing respectively.
-Player.prefab. Added the trail renderer for the dash trail.

How to test: load up the sprint 3 scene: When you dash a trail should appear behind you that matches your character's color.


IanGRap:
will do

PR #95, opened 2019-03-04
Added squash and stretch animation to block push

IanGRap:
Contributors: Ian

Goal: Add a squash and stretch animation to blocks when you push them to give the game more juice.

Important Files: 
- Block Mesh.controller and SquashAndStretch.anim: The animation and control of the block squash and stretch. 
- BlockPusher.cs: Calls the squash and stretch animation when pushed. Also you can tweak squashAndStretchOffset to set the delay of when the animation should play after the blocks is pushed. This delay is added recursively to multiple pushed blocks. It makes the cool wave effect.
- BlockCollisionDelegator.cs and BlockInput.cs: Call push with no delay for the animation.

How to test: Load up the sprint 3 scene and push some blocks. The blocks should have a subtle squash and stretch animation when they move. There should also be a delay that makes a wave effect.


IanGRap:
done

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -36,11 +40,18 @@ private void Awake()
...
     }
 
-    public bool Push(Vector3 direction)
+    private IEnumerator SquashAndStretchDelay(float delay)
+    {
+        yield return new WaitForSeconds(delay);
+        _animator.Play("SquashAndStretch");
+    }
+
+    public bool Push(Vector3 direction, float animationDelay)
```
AustinKellar: Since you are passing in 0f in all cases here, can we set a default value of 0 for animation delay? I.E

public bool Push(Vector3 direction, float animationDelay = 0f)

IanGRap: yeah I can make that change

PR #96, opened 2019-03-05
Block spawning intensity

IanGRap:
Contributors: Coulter, Ian

Goal: Make the spawn rate of blocks dynamic to keep various states of the game interesting.

Important Files: 
- Spawner.cs: Tons of changes. We made it an instance. It tracks the count of blocks in the scene. When the coroutine for the spawn restarts it calculates the timer. The timer is influenced by the number of blocks in the scene, the number of players in the scene and the width of the arena. When these values go down the spawn rate increases. 
- BlockDamager.cs: Updates the spawner's block count when a block is spawned and destroyed.
- MatchManger.cs: Has a public property for the current player count.
- WallManager.cs: Has a public property for the minimum arena width.
- BlockClearer.cs: Calls block break function when clearing a lane.

How to test:
Load up the sprint 3 scene and play. You may need to revert prefabs in the scene like the match manager and block spawner. You can tweak different values on the spawner. The influence values on the spawner say how much each respective aspect of the game influences the spawn rate. The influences need to add up to 1 because math. The high block count is the number the of blocks that spawner sees as too many and will slow down spawning as a result. These factors influence where the timer value will be inbetween the fast timer and slow timer. 

While playing. the spawn rate should increase when there are less players, the arena is smaller and there are less blocks in the scene. It should slow down in all the opposite cases.

Notes:
- We ran into a bug where blocks stopped spawning but we think we might've fixed it.
- We didn't end up adjusting the block gravity speed since we changed so many other values and want to see how those respond to playtesting.


Assets/Scripts/BlockSpawner/Spawner.cs
```diff
@@ -132,13 +218,17 @@ private float RoundToNearestInterval(float value, float interval)
         }
     }
 
-    private IEnumerator SpawnTimer(float timer)
+    private IEnumerator SpawnTimer()
     {
-        yield return new WaitForSeconds(0.5f);
+        yield return new WaitForSeconds(0.1f);
```
AlfredoCuevas: Because we are using this so much it might be beneficial to just make a variable out of it so that we aren't constantly making new instances of it. 
Ex) WaitForSeconds w = new WaitForSeconds(0.1f);

IanGRap: oh I didn't know you could do that. I'll check it out

AlfredoCuevas: yea i think it would help for inside the loop. I'm not sure how often that is being called though

PR #98, opened 2019-03-06
Created deathblock variant that destroys anything it touches.

IanGRap:
Contributors: Ian, Coulter

Goal: Create a new block variant that kills or destroys all blocks and players in its path, and then immediately gets destroyed on collision with the ground or walls.

Important Files Changed:

- Spawner.cs: Added deathblock to list of randomly spawned formations.

- Deathblock scripts: Created movement and collision scripts unique to the deathblock. It will destroy any players or blocks with which it collides.

- SquishDetector.cs: made player death function publicly accessible.

How to Test: Load sprint 3 scene and wait for a deathblock to spawn. It has a distinctive glowing red appearance as a visual warning. Let it destroy the blocks under it, and try to interact with it as the player. It should simply move down at a constant speed and destroy all blocks and players it touches until it hits the ground, at which point it will break.


Assets/Scripts/Blocks/DeathBlockMovement.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class DeathBlockMovement : MonoBehaviour
+{
+    [SerializeField]
+    private float _fallSpeed;
+
+    private Rigidbody _deathBlockRB;
```
AustinKellar: Our convention has been to just call this _rigidbody. Can we stick with that?

Issue #101, opened 2019-03-07
Sometimes you can fall in between adjacent blocks


AustinKellar:
Reproduce case:
Fall from a far distance directly between two blocks. Occasionally, you will fall in-between them and get stuck inside blocks.

Issue #83, opened 2019-03-01
Sometimes, you don't die if a block falls on you.


egomish:
How to replicate:
- Find two adjacent blocks that are falling
- Jump or dash above them so that you're directly between them
- Finish falling

Issue #29, opened 2019-02-08
Sometimes, you die if you dash into a block


AustinKellar:
How to replicate:
- We haven't found a reliable way to replicate this bug, all we know is that sometimes when you dash into a block, you will die.
- We have only seen this happen with horizontal dashing so far.

Notes:
- This bug has reappeared. It appears to only happen when you dash at the point between to adjacent blocks. Likely just needs another tweak of the squish values.

AustinKellar:
Closed by PR #37

PR #102, opened 2019-03-07
Bugfix #83, #29, #101

AustinKellar:
Contributors: Austin

Goal:
- Fix the following issues
    - Falling in-between two adjacent blocks
    - Sometimes when a block falls on the player's head, it will bob up, then back down before it squishes the player
    - Sometimes, the player will randomly die when dashing into blocks (not sure if this one is better. It's hard to tell)

What changed:
- Re-calibrated the location of the player's squish colliders
- Increased the size of the player prefab
- Rather than using the scale of the player objects in the starting block configuration prefabs, use the scale of the player prefab (I found this to greatly reduce the rate of players falling between blocks, because the localScale is consistent)
- Added an extra box collider near the feet of the player prefab (I found this to also help with the player falling between blocks)

How to Test:
- Load up the Round System scene and try to reproduce the bugs
- Please let me know if you do and if the rate has seemed to improve / get worse


AustinKellar:
I agree that everyone should test if possible.

PR #110, opened 2019-03-09
Added extra spawn point for 3-player matches.

JamesCoulter:
Contributor: Coulter

Goal: Spread out spawn points when only three players are in the match.

Files Changed:

- SceneCreator.cs: Expanded player re-position function to account for 3 players.

- All starting structure prefabs: Added a special spawn point to the middle area of each prefab that is only used if 3 players are in the game.

How to Test: Start a game with 2, 3 or 4 players (test all numbers a few times). The players should always spawn spread out appropriately depending on the active number.


JamesCoulter:
Resolved merge conflict, ready to be reviewed now

Assets/Scripts/Managers/SceneCreator.cs
```diff
@@ -108,13 +110,24 @@ private Transform ChooseStartStructure()
...
+            return;
+        }
         if(MatchManager.Instance.PlayerCount == 2)
         {
             MatchManager.Instance.AlivePlayers.First().transform.position = _playerTransforms.First();
             MatchManager.Instance.AlivePlayers.Last().transform.position = _playerTransforms.Last();
         }
+        else if(MatchManager.Instance.PlayerCount == 3)
+        {
+            _middleSpawn = GameObject.FindGameObjectWithTag("Spawn");
```
AlfredoCuevas: If possible I would like to see this changed to something that is less computationally expensive. Possibly by looking specifically inside of the children of the Starting Structure.

JamesCoulter: I changed it so that _middleSpawn just looks through the structure children to find the object. I also noticed that in scenecreator.cs the _structurePrefab field was unused, so I went through the SpawnStructure() method and replaced the local structure variable with _structurePrefab (so that I could access the selected structure outside of that method). Unfortunately I can't test the changes now. If you are able, can you see if it still works?

AlfredoCuevas: I tested with 3 controller and the spawning seemed to work the same. Looks good to me

JamesCoulter: Cooleo, I'll merge it in then

PR #111, opened 2019-03-09
New Player Select Lobby!

AustinKellar:
Contributors: Austin

Goal:
- Port the new player select lobby into existing game

What Changed
- All new code can be found in the PlayerSelect folder

How to Test:
- Load scene New Lobby and make sure functionality works as expected
- Load New Lobby Match Scene and make sure that players are spawned in correctly with unique colors
- Test in standalone build


Assets/Scripts/Player Selection/BlockSpawner.cs
```diff
@@ -0,0 +1,138 @@
+﻿using System;
```
IanGRap: I think the name of this script is confusing given the block spawner in the main game

Assets/Scripts/Player Selection/BlockSpawner.cs
```diff
@@ -0,0 +1,138 @@
...
+        if (_spawnStack.Contains(position))
+        {
+            return;
+        }
+
+        Destroy(block);
+
+        ParticleSystem particleEmitter = Instantiate(_despawnParticleEffect);
+        particleEmitter.transform.parent = gameObject.transform;
+        particleEmitter.transform.position = blockTransform.position;
```
IanGRap: you can call Instantiate with things like position if you want to reduce your lines some

Assets/Scripts/Player Selection/SpecialBlockGravity.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class SpecialBlockGravity : MonoBehaviour
+{
+    private void Start()
+    {
+        Physics.gravity = new Vector3(0, -75f, 0);
```
IanGRap: won't this also change physics for the player?

Assets/Scripts/Player/SquishDetector.cs
```diff
@@ -75,6 +85,11 @@ private bool SquishedFromAbove()
 
     private bool SquishedFromTheSide()
     {
+        if (MatchManager.Instance == null)
+        {
+            return false;
+        }
+
```
IanGRap: There is a ton of new code being introduced here with only 1 person having written it. I'd be pretty worried if we continued this trend in the future.

Assets/Scripts/Player Selection/BlockSpawner.cs
AustinKellar: Sounds good. How about PlayerSelectBlockSpawner

Assets/Scripts/Player Selection/BlockSpawner.cs
AustinKellar: I'll look into that

Assets/Scripts/Player Selection/BlockSpawner.cs
IanGRap: That works for me

Assets/Scripts/Player Selection/SpecialBlockGravity.cs
AustinKellar: Nope, gravity is disabled for every object except these blocks. If you want, I can write custom gravity for this.

Assets/Scripts/Player/SquishDetector.cs
AustinKellar: Would you prefer to not merge this? We can scrap the idea if you're uncomfortable with it. Or rewrite it. Whatever works

Assets/Scripts/Player Selection/SpecialBlockGravity.cs
IanGRap: oh yeah I forgot. This should be fine then.

Assets/Scripts/Player/SquishDetector.cs
IanGRap: I don't think we should scrap it but this also probably shouldn't be a trend we continue in the future. I think the main thing now is just to make sure every programmer reviews it

IanGRap: There is a lot of good stuff in here and it would be a waste to let it all go

Assets/Scripts/Player Selection/ControllerSignIn.cs
```diff
@@ -0,0 +1,101 @@
...
+            PlayerSelectUIManager.Instance.RemovePlayer(_index);
+            if (!PlayerSelectUIManager.Instance.ReadyToStartMatch)
+            {
+                _playerSpawner.Despawn(_index, true);
+            }
+        }
+        if (state.Back)
+        {
+            Application.Quit();
+        }
```
IanGRap: I feel like this should only quit if the player does not have a character in the scene.

Assets/Scripts/Player Selection/PlayerSelectBlockSpawner.cs
```diff
@@ -0,0 +1,158 @@
...
+        Material materialToRemove = _materialsInUse.FirstOrDefault(m => m.color == playerMaterial.color);
+        if (materialToRemove)
+        {
+            _materialsInUse.Remove(materialToRemove);
+        }
+        _paintBlockPositions.RemoveAll(m => m == newPaintBlockPosition);
+    }
+
+    public Material GetPreferredOrRandomMaterial(Material preferredMaterial)
+    {
```
IanGRap: Could you make this function have an override and if there is no argument you get the random material?

Assets/Scripts/Player Selection/PlayerSelectBlockSpawner.cs
```diff
@@ -0,0 +1,158 @@
...
+                !PlayerSelectUIManager.Instance.ReadyToStartMatch && 
+                _materialsInUse.Count < Materials.Instance.availableMaterials.Length &&
+                _spawnCount >= 4 &&
+                _playerSpawner.DeadPlayerCount == 0)
+            {
+                SpawnPaintBlockFromStack();
+            }
+            yield return new WaitForSeconds(_paintBlockRespawnTimer);
+        }
+    }
```
IanGRap: does this mean you can't change your color if the match can be started?

AustinKellar: Kinda. When the "Start Match" Screen pops up, blocks won't spawn (just so people don't die unexpectedly by spawning blocks. But if the cancel out of the screen, they start spawning again.

JamesCoulter:
A couple things right off the bat:
1) Player point lights aren't set to the player's material color. They're always green
2) I think holding B to exit the lobby is fine, but I think there should be a visual indicator around the B button in the UI that indicates how long the player needs to hold the button to exit. This issue is super minor, just a thought I had. 

Still need to go through all of the actual code, so I'll focus on that now.

PR #112, opened 2019-03-09
Block health fix

JamesCoulter:
Contributor: Coulter

Goal: Change block health mechanic so that health is only decremented after the block is done being pushed.

Files changed:

- BlockPusher.cs: added bool property that gets set to true when IsPushed gets set to false. Also added private _isPushed variable to be accessed by the IsPushed property.

- BlockDamager.cs: added update() in order to check for IsPushed value going from true to false and, if true, decrement health.

How to Test: Start the scene and push blocks. Their color shouldn't change and they shouldn't break until after the current push is completed.

Notes: May conflict with changes from other branches that haven't been merged yet.


JamesCoulter:
Somehow my 3 player spawn changes and my block health changes have been combined into the same branch or same PR or something, even though they were done in completely separate branches and neither was ever merged into the other

IanGRap:
You probably did checkout -b from the other branch when you made this branch. Doing that carries all your changes from the previous branch. Just make sure the other branch gets merged in before this one

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -21,7 +23,34 @@ public class BlockPusher : MonoBehaviour
...
+    public bool ValueHasChanged
+    {
+        get
+        {
+            return _valueHasChanged;
+        }
+        set
+        {
+            _valueHasChanged = value;
+        }
```
IanGRap: you can just right this as { get; set;}

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -9,6 +9,8 @@ public class BlockPusher : MonoBehaviour
 {
     [SerializeField]
     private float _pushDistance = 1f;
+    private bool _isPushed;
+    private bool _valueHasChanged;
```
IanGRap: This variable name is a little unclear.

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -21,7 +23,34 @@ public class BlockPusher : MonoBehaviour
 
     public bool IsPushed
     {
-        get; set;
+        get
+        {
+            return _isPushed;
```
IanGRap: Why do we need _isPushed? Can't we just return IsPushed? Can't we set IsPushed = value?

JamesCoulter: I tried that, and using IsPushed causes an endless loop of accessing the property and crashes Unity. I learned that unless you're using an auto-implemented property, like for example:
`public int Property{ get; set; }`
You are required to have a backing field, in this case I did _isPushed. I could maybe look into doing all this in a way that doesn't require the _isPushed field, this is just the first and most common method I saw online for keeping track of a value change.

IanGRap: Ahhh interesting. I wasn't aware

Assets/Scripts/Blocks/BlockPusher.cs
JamesCoulter: Changed it to BlockPushEnded. Let me know if that's better

Assets/Scripts/Blocks/BlockPusher.cs
JamesCoulter: Stupid of me. Done

Assets/Scripts/Blocks/BlockDamager.cs
```diff
@@ -48,6 +58,7 @@ public void DecrementHealth()
             _pushCount++;
             _blockMat.color = new Color(_blockMat.color.r - _percent, _blockMat.color.g - _percent, _blockMat.color.b - _percent, _blockMat.color.a);
             _blockRenderer.materials[0] = _blockMat;
+            _blockPusher.ResetFlag();
```
AlfredoCuevas: ResetFlag seems a bit unclear to me. Maybe be more explicit and say ResetBlockPushEnded(). There are probably better names but ResetFlag seems a little generic

Assets/Scripts/Blocks/BlockDamager.cs
```diff
@@ -40,6 +42,14 @@ public void Break(bool playSound = true)
         Instantiate(_breakEffect, transform.position, Quaternion.identity);
     }
 
+    private void Update()
```
AlfredoCuevas: Update should be moved before the public Break() function inside of the script.

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -20,6 +21,26 @@ public class BlockPusher : MonoBehaviour
...
+            }
+            _isPushed = value;
+            if(!_isPushed)
+            {
+                BlockPushEnded = true;
+            }
+        }
+    }
+
+    public bool BlockPushEnded
```
AlfredoCuevas: BlockPushEnded should probably have a private set; Otherwise what is the point of having the ResetFlag() function when you can just directly make the variable false from BlockDamager()

PR #113, opened 2019-03-11
Cross bomb block

AlfredoCuevas:
Contributors: Ian, Alfredo

Goal: Create a Bomber Man style block that explodes in 4 directions killing / destroying everything in the way. 

Files Changed: 
* BlockDamager/BlockPusher.cs: certain functions have been made virtual to allow for overriding
* BlockExploder.cs : Derives from BlockDamager class. This adds the main functionality for the bomb block falling, exploding, and playing audio
* BombPusher.cs: Derives from BlockPusher class. This calls the BlockExploder's Detonate() when pushed
* CrossBombBlast.cs: This gives functionality to the explosion part of the bomb which is instantiated when the bomb detonates
*  DestroyOnTimer.cs: Used to destroy the cross explosion. 
* Spawner.cs : Instead of checking for a tag = "Block" we created a list of tags so that multiple types of blocks can be included in the check.
* BlockAdjacencyTracker / PlayerAdjacecyTracker/ Wall.cs : Now checks the list of tags vs just "Block"
* UIMatchManager.cs: Now has a function to stop the bomb's ticking sound when the players leave the match. 
* Prefabs have been added for the CrossBombBlock and the CrossExplosion
* AudioManager Prefab: adjusted to add the new sounds for the block

**How To Test**
Run the sprint4 playerselect scene. Play the game. The block should spawn every once in a while. Make sure to notice if the sounds are playing correctly. The ticking sound should start playing when the bomb spawns out of view. Then the bomb will turn red and explode. The ticking will stop only if there are no bombs in the scene. 
Also make sure to transition to leave the match when the bomb is in the arena to make sure that none of the sounds are messed up. 

* Note: No proper art has been added for the block yet. We discovered a small issue with the audio manager fading function.


Assets/BlockExploder.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using DG.Tweening;
+
+public class BlockExploder : BlockDamager {
```
AustinKellar: curly brace on next line plz

Assets/CrossBombBlast.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class CrossBombBlast : MonoBehaviour {
```
AustinKellar: curly brace on next line plz

Issue #106, opened 2019-03-08
Win screen defaults sometimes


IanGRap:
If the remaining players die very close to one another but the last player to die won the entire match, the win screen will default to a white skeligoo picture with a Oozing Orange as its name

PR #117, opened 2019-03-13
Bugfix #106

AustinKellar:
Contributors: Austin

Goal
- Fix the bug where sometimes, when two players die in quick succession, the game displays Oozing Orange as the winner with a _white_ skelegoo icon (even if Oozing Orange is not in the game)

What Changed:
- In MatchManager.cs, set the winning player when elements are removed from the _alivePlayers array. This guarentees that we do not destroy the game object before trying to read from it.
- In UIMatchManager.cs, check if WinningPlayer != null to pull up game over screen, instead of checking if AlivePlayers.Count == 1

How to Test
- Test in development first
- Go to the New Lobby Scene and start a match
- Get to Game Point (someone has 2 points and will win if they win another match)
- Create this scenario (where the winning player is on the top)
    
image

- You should at this point see the error (if you are in development)
- Pull this branch, create the same scenario and see if the bug still pops up


AustinKellar:
Really? Noooo! Okay. I'll look at it more tomorrow.

AustinKellar:
I took another crack at it. Changed the way the GameOverUI reads the Material of the player. It gets the info from ActivePlayers rather than MatchManager, since MatchManager's info changes constantly

PR #127, opened 2019-03-14
First release build

IanGRap:
Contributors: Reed, Ian

Goal: Put together the scenes for the first playable release.
Important Changes:
- Tweaked particles for player death and block breaks
- Added post processing in
- Added in new floor
- Redid background details
- Added in new walls

How to test: Load up the lobby scene and the first playable scene. Everything should look good and play right. Also test in build.


Assets/CameraGoo.cs
```diff
@@ -38,11 +38,10 @@ private void KeepOnlyOneInstance()
         if (Instance == null)
         {
             Instance = this;
-        }
-        else
+        } else
```
AlfredoCuevas: put this down

PR #132, opened 2019-03-22
Lobby Import w/ Title Screen

AustinKellar:
Contributors: Austin

Goal: 
- Import the lobby changes that I made in CM 121.
    - Title screen before the color selection lobby
    - Checkmarks above the heads of ready players
    - Title that says "Pre-Game Lobby"

What Changed:
- New files
    - LobbyStateManager (keeps track of the current state of the lobby )
    - UITitleScreen
    - UITitleScreenInput
- PlayerSpawner, PlayerSelectBlockSpawner (added a reset function that is called when you return to title screen)
- PlayerSelectUIManager (added functions to enable / disable checkmark image above player)
- Player prefab (added image of checkmark as a child)

Notes:
- Let me know your thoughts of the post-processing on the title screen. IMO, it looks weird.
- I dabbled with changing the controls to be entirely on start and back and I did not like the feeling. I think its wise to have a meeting about it before implementing that change.


Assets/Scripts/Player Selection/LobbyMenuInput.cs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private void CaptureInput()
+    {
+        CloseMenuPressed = _gamePadState.Pressed(CButton.B);
+
+        bool pressedDown = _gamePadState.GetAxis(CAxis.DY) == -1f || _gamePadState.GetAxis(CAxis.LY) == 1f;
+        DownButtonPressed = ((pressedDown && !_downPressedLastFrame) || Input.GetKeyDown(KeyCode.DownArrow));
+        _downPressedLastFrame = pressedDown;
+
```
IanGRap: Does carbon Input not have a function for OnButtonDown (something that only returns true if the button was pressed this frame?)

Assets/Scripts/Player Selection/LobbyStateManager.cs
```diff
@@ -0,0 +1,88 @@
...
+        _destination = _startingPosition = _camera.transform.position;
+        _destinationRotation = _startingRotation = _camera.transform.eulerAngles;
+    }
+
+    private void FixedUpdate()
+    {
+        _camera.transform.position = Vector3.Lerp(_camera.transform.position, _destination, 0.05f);
+        _camera.transform.eulerAngles = Vector3.Lerp(_camera.transform.eulerAngles, _destinationRotation, 0.05f);
+    }
+
```
IanGRap: Since none of these are physics calculations the Unity convention would usually be to put this in Update(). Doesn't really matter to me though.

Assets/Scripts/Player Selection/PlayerSelectBlockSpawner.cs
```diff
@@ -82,12 +102,32 @@ private void SpawnPaintBlockFromStack()
...
+
     private IEnumerator RespawnPaintBlocks()
     {
         while (true)
         {
-            if (_blocksToRespawn.Count > 0 && 
-                !PlayerSelectUIManager.Instance.ReadyToStartMatch && 
+            if (_blocksToRespawn.Count > 0 &&
+                LobbyStateManager.Instance.State == LobbyStateManager.LobbyState.ColorSelect &&
+                !PlayerSelectUIManager.Instance.ReadyToStartMatch &&
```
IanGRap: This if statement is getting kinda unreadable. Can you break all these checks into bool variables so its easier to know what they are each doing?

Assets/Scripts/Player Selection/LobbyStateManager.cs
AustinKellar: Will do.

Assets/Scripts/Player Selection/LobbyMenuInput.cs
AustinKellar: Yes it does but it can't be used on the analog stick. The analog sticks are axes, so gotta do it manually :/

Assets/Scripts/Player Selection/PlayerSelectBlockSpawner.cs
AustinKellar: I'll see what I can do

Assets/Scripts/Player Selection/LobbyMenuInput.cs
IanGRap: ahhh gross but necesary then

Assets/Scripts/Player Selection/PlayerSelectBlockSpawner.cs
AustinKellar: Do you basically just want a bunch of one-line properties / functions for each of these checks? So that they have names?

IanGRap: That's what I was thinking.

AustinKellar: Okay.

PR #135, opened 2019-04-03
Small spring tweaks

IanGRap:
Contributors: Ian

Goal: Make various tiny quality of life improvements while we wait for sprint planning to be finished. Addresses issues 129, 109 and 92.

Changes: 
- SquashAndStretch.anim: Increased the amount of squash and stretch on the animation when you push blocks to make it more obvious and more juicy
- BlockExploder.cs: Refactored public static value into private value with access functions. Changed detonation color tween from red to white to make it more visible. Adds self to camera tracking temporarily on detonation to be more easy to track by players. Calls warning blast and wind up sound on detonation. Shakes camera on explosion.
- CrossBombBlast.cs: Refactored tag check so bomb can blow up other bombs.
- ExplosionWarningController.cs: Creates a growing cross warning shape right before the bomb explodes to alert players.
- CrossExplosion.prefab: Decreased the size of the cross bomb explosion slightly to make it a little less punishing.
- [BlockSpawner].prefab: Changed spawning to iterate over each player instead of spawning randomly over each player to make the game a bit more fair.
- Block Clearer.prefab: Increased the shake maginitude for the camera when a lane clears to make the shake slightly more noticable.
- [AudioManager].prefab: Added bomb wind up sound and footstep sounds.
- DeathBlockCollision.cs: Changed tag check so death block can detonate bombs.
- Cameramanager.cs: Added new value that lets designers change the pan speed of the camera in editor. Set the current pan speed to double what it was previously to make it more snappy. Made it so shake time doesn't stack if multiple things call shake at once. Also added default shake values.
- CharacterAnimationEvents.cs: A script on the player mesh that has functions animations can call. Right now there is only a footstep sound call. This is muted currently because I don't think it sounds good, but we can tweak it later.
- SquishDetector.cs: Getting squished shakes the camera.
- Walls.cs: The emoji colors change from green to red right before the walls compact.
- WallManager.cs: Walls compacting shake the camera.

How to test: Load up a build a see if the above changes are working. Do any of them seem like negative additions?

Notes: Footstep sounds are currently muted because I don't think they sound right.


Assets/Scripts/Managers/UIMatchManager.cs
```diff
@@ -194,6 +194,6 @@ private void StopSFX()
         }
 
         //Reset bomb count before Loading
-        BlockExploder.BombsInScene = 0;
+        BlockExploder.SetBombCount(0);
```
AustinKellar: Unless this function is called somewhere to set the bomb count to something other than 0, I would like to see this renamed to BlockExploder.Reset(), with no arguments.

PR #141, opened 2019-04-07
Issue 138: Mac Build not working

IanGRap:
Contributors: Ian

Goal: Make the game work on Macs. 

Important Changes: Changed the shader settings for macs to avoid error.

How to test: Mac version shouldn't have major errors and should run with all content.


AustinKellar:
I'm not sure how to test this. Maybe we can watch you load it up on your mac?

IanGRap:
It works on my mac. I think we will need Emma or another team member with a mac to test and confirm. This would be an unorthodox pr

egomish:
- The old controller bug, where one player can't move unless the other player is moving on their controller, has resurfaced.
- The game is a little laggy at ultra graphics settings.
- The dash trail is black, rather than colored.
Unless we figure out and fix the controller bug, I don't think it's a good idea say we support Mac.

PR #142, opened 2019-04-08
Dash bounce

AustinKellar:
Contributors: Austin

Goal:
- Add a bounce back when the player collides with objects during a dash

What Changed:
- I added collision events to PlayerDash, which adds a small force when the player collides during a dash
- I added an option to PlayerMovement which disables movement while the player is bouncing

Customization:
- PlayerDash
    - BounceOnDashCollision (enables / disables the feature)
    - BounceBackForce (the force modifier of the bounce)
    - BounceBackUpwardTravel (how far upwards the player travels in a bounce)
    - BounceResetTimer (the amount of time before the player is considered to be not bouncing)
    - BounceBackLayerMask (controls which objects the player can bounce off of)
- PlayerMovement
    - MoveWhileBouncing (disables player movement while bouncing [this is where the BounceResetTimer is useful])

How to Test:
- Load up both the lobby and the match scene and make sure you bounce off of the objects in the LayerMask (by default its everything but the floor)
- Test all options that I laid out in Customization


Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -47,11 +63,14 @@ public class PlayerDash : MonoBehaviour
     private float _previousDashTime;
     private bool _groundedSinceLastDash;
     private float _tolerance;
+    private bool _isBouncing;
```
IanGRap: Why does _isBouncing need to be its own variable? Can we just have the property IsBouncing with a private set?

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -104,6 +133,25 @@ private void FixedUpdate()
...
 
+    private void BounceBackOnDash(Collision collision)
+    {
+        if (_bounceOnDashCollision && 
+            IsDashing && 
+            !_isBouncing && 
+            _bounceBackLayerMask == (_bounceBackLayerMask | (1 << collision.gameObject.layer)))
+        {
+            Vector3 bounceDirection = new Vector3(DashDirection.x * -1, _bounceBackUpwardTravel, 0);
+            _rigidbody.AddForce(bounceDirection * _bounceBackForce, ForceMode.Impulse);
```
IanGRap: This calculation is a little confusing to me. Can we make it that instead of having a variable of _bounceBackForce, we have _bounceBackUpwardTravel and _bounceBackHorizontalTravel? Then when you add the force you just use those 2 values in a vector and don't worry about the force. While testing, I was having trouble setting the values so that skeligoo would bounce up more but horizontally less.

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -104,6 +133,25 @@ private void FixedUpdate()
...
+            _isBouncing = true;
+            Invoke("ResetBounce", _bounceResetTimer);
```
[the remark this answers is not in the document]
AustinKellar: Yep

Assets/Scripts/Player/PlayerDash.cs
AustinKellar: Done

Assets/Scripts/Player/PlayerDash.cs
AustinKellar: Done

Assets/Scripts/Player/PlayerDash.cs
AustinKellar: I had it as a private field so that I could easily see its value in Debug mode but that is no longer needed. I changed it to match your description

IanGRap: In the unity editor you can set it to debug so you can see private variables

AustinKellar: Right, that's the mode I'm talking about. To my knowledge, it does not work with properties, only basic fields

AustinKellar: Oh, it seems I was wrong about that. It does work with the property. In any case, I changed it

AustinKellar:
Oh goodness, that was a dumb mistake. Can't believe I didn't try that. I'll get on it.

IanGRap:
Also it might be easier if when a block gets pushed it goes back and tells the player that pushed it that it got hit

AustinKellar:
It is indeed a problem with this PR. I added a new layer to paint blocks so that the player could bounce off of them but forgot to add that layer to the PlayerGrounded layer mask. I just updated it. Can you try it again?

AlfredoCuevas:
> It is indeed a problem with this PR. I added a new layer to paint blocks so that the player could bounce off of them but forgot to add that layer to the PlayerGrounded layer mask. I just updated it. Can you try it again?

Looks good now

Issue #140, opened 2019-04-06
Players crushed by lane clearing


IanGRap:
When lanes clear, players have a tendency to fall under blocks and get squished. This feels too punishing

PR #145, opened 2019-04-11
Players crushed by lane clearing Issue #140

IanGRap:
Contributors: Ian

Goal: Make it so players couldn't accidentally fall beneath blocks when a lane clears.

Important Files: 
- PlayerOverlapTracking.cs: Keeps track of players within a trigger box (one that goes just above the bottom row) and can apply a force to all these players.
- BlockClearer.cs: Sets the size of the overlap tracker and calls the push function when a lane clears. Change PlayerPushForce to change the force applied to players when a lane clears.

How to test: Load up a match. When the bottom row of blocks clears, try to see if you can dart to the bottom floor and get crushed by the blocks moving to the bottom floor. You should be unable.


AustinKellar:
New changes look good to me. I saw no issues.

PR #147, opened 2019-04-13
Long and Tall Block Shapes

JamesCoulter:
Contributors: Coulter, Alfredo

Goal: Create tall block (1x3) and long block (3x1).

Files Changed:
- GetFormationLength.cs: Simple script to pass the shape's width value to the spawner to avoid spawning outside or on top of the walls.
- Spawner.cs: Added tall and long block prefabs and formations to the spawn formation list. Also adjusted spawn position shift calculations to account for width values passed by GetFormationLength.cs.
- BlockClearer.cs: Added _overlapCount variable to keep track of how filled the bottom row is (instead of using the list count).
- SquishDetector.cs: Removed unneeded bool that was messing up all block squishing in the downward direction.

How to test: 
- Go into "First Release 1" scene
- Run the game
- Interact with the shapes, observe their interactions with each other (and with special blocks)

Notes: There's a camera manager error popping up occasionally, not our fault. Shout out to Austin for fixing his fuck-up.


Assets/Scripts/Blocks/GetFormationLength.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class GetFormationLength : MonoBehaviour {
```
AustinKellar: Curly brace is on the wrong line

AustinKellar: Can we just call this class BlockFormation? I feel like this name sounds like a function name rather than a class name. If we call it BlockFormation, we can also include other data in here if we need it down the line.

Assets/Scripts/Environment Actions/BlockClearer.cs
```diff
@@ -51,7 +52,13 @@ private void UpdateDelegation()
         _overlapBlocks.RemoveAll(obj => obj == null);
         if (_clearOnRowFilled)
         {
-            if (_overlapBlocks.Count == WallManager.Instance.ArenaWidth)
+            _overlapCount = 0;
+            foreach(GameObject obj in _overlapBlocks)
+            {
+                if (obj.tag == "LongBlock") _overlapCount += 3;
```
AustinKellar: Please put curly braces on these if statements. That is part of our style guide

AustinKellar: I'm somewhat uncomfy with this += 3 being a hard coded value, honestly. What if we have a long block with 4 blocks in there later?

PR #151, opened 2019-04-16
Floor grid

AlfredoCuevas:
Contributors: Alfredo Cuevas

Goal: Make it so that the floor spawns in grid and is stored in a 2d array

What changed: 
* Made prefabs for the floor cubes
* Made a prefab for the floor with the spawner script
* Floor.cs: Creates the floor on awake. Blocks start spawning at the location of the spawnpoint. Adds the objects to a 2D array

How to test: 
Go into either the First Release scene or the scene that I made (Procedural Floor). Run the application. the floor should be immediately spawned on awake.


Assets/Scripts/Floor.cs
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    if (j % 2 == 0)
+                    {
+                        cube = Instantiate(_greenBlockPrefab, gameObject.transform);
+                    }
+                    else
+                    {
+                        cube = Instantiate(_purpleBlockPrefab, gameObject.transform);
+                    }
+                }
```
IanGRap: You could maybe simplify this bit of code by seeing if i + j % 2 == 0. You would need less if else blocks that way

PR #152, opened 2019-04-17
Respawn as block

AustinKellar:
Contributors: Austin, James

Goal:
Create a system where players have a chance to respawn after they die. 

What Changed:
- BlockDamager.cs
    - Added logic to the Break() function so that the block spawns a player if it is a RespawnBlock and it squished another player before being destroyed
    - In Update, if the block is a RespawnBlock and FallingPathBlocked() is true, then destroy the block
- Spawner.cs
    - Added a list (which functions as a queue) to store the players waiting to be spawned as a block. 
    - When there is a player waiting to respawn, regular block spawning is paused while the Spawner spawns respawn blocks
- MatchManager.cs
    - Added logic for _undoing_ a player death. I.E. added a function to add a player back into _alivePlayers
- PlayerSquishDetector.cs
    - When a player gets squished, if they were squished by a respawn block, tell the respawn block that it killed a player, so that it knows to spawn in a new one.
- PlayerSpawnBeam.cs
    - Contains logic for the beam of light when a player spawns in
- BlockInput.cs
    - Listens for controller input and moves respawn blocks
- CameraTracker.cs
    - Controls when the respawn block gets added to the camera manager

How to Test:
- Load up the match scene and die. You should see a respawn block spawn in after ~3 seconds
- Make sure you can control the respawn block with controllers
- Make sure that respawn blocks can squish players
- Make sure respawn blocks CANNOT push other blocks
- Make sure respawn blocks are destroyed once they stop falling
- Make sure that if a respawn block kills a player, it spawns in a new one of the correct color and controller input
- Change the Respawn Delay value on the Block Spawner. This changes the minimum delay time between player death and when they respawn as a block
- Change the Camera Track Y Coordinate value on the CameraTracker component of the respawn block prefab. This field describes the height at which the camera starts tracking respawn blocks

PRO TIP:
- This feature is way easier to test if you reduce the block spawning rate. Players don't tend to live long enough with the current spawn rate, at least when nobody is controlling them.


Assets/Scripts/BlockSpawner/Spawner.cs
```diff
@@ -300,4 +338,18 @@ private IEnumerator SpawnTimer()
...
+        public Material material;
+        public float queueTime;
+
+        public RespawnBlockInfo(PlayerIndex i, Material m)
+        {
+            index = i;
+            material = m;
+            queueTime = Time.time;
+        }
+    }
```
IanGRap: We probably don't have the time, but in an ideal world Spawner.cs would probably need a refactor since its getting so long. Maybe RespawnBlockInfo can go somewhere else? Maybe its own file?

Assets/Scripts/Blocks/BlockDamager.cs
```diff
@@ -4,6 +4,12 @@
 
 public class BlockDamager : MonoBehaviour
 {
+    [SerializeField]
```
IanGRap: I think these changes should all go in a different file that inherits from BlockDamager.cs. Most blocks don't need this functionality. You can look at BlockExploder as reference (it also inherits from BlockDamager).

Assets/Scripts/Blocks/CameraTracker.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            if (transform.position.y <= _cameraTrackYCoordinate)
+            {
+                CameraManager.Instance.AddTrackedTransform(gameObject.transform);
+                _hasBeenAdded = true;
+            }
+            yield return _wait;
+        }
+    }
+}
```
IanGRap: This file seems unnecessary. I believe the camera manager has a function for adding transforms on a delay. If that isn't the case, I feel that that functionality should be added to the camera manager

Assets/Scripts/Blocks/PlayerSpawnBeam.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            other.gameObject.GetComponent<BlockDamager>().Break();
+        }
+    }
+
+    private void DestroySelf()
+    {
+        Destroy(gameObject);
+    }
+}
```
IanGRap: You could use the destroy on timer script on this gameobject instead of putting the functionality in this script. Your call though

Assets/Scripts/Player/PlayerBounceBack.cs
```diff
@@ -45,7 +46,8 @@ private void OnCollisionStay(Collision collision)
     {
         if (_playerDash.IsDashing &&
             !IsBouncing &&
-            _bouncyEnvironmentObjects == (_bouncyEnvironmentObjects | (1 << collision.gameObject.layer)))
+            (_bouncyEnvironmentObjects == (_bouncyEnvironmentObjects | (1 << collision.gameObject.layer)) || 
+             collision.gameObject.tag == "RespawnBlock"))
```
IanGRap: OnCollisionEnter and OnCollisionStay have this same functionality. Can we move this code into its own function?

Assets/Scripts/Blocks/BlockDamager.cs
AustinKellar: Sounds like a good idea

Assets/Scripts/Blocks/CameraTracker.cs
AustinKellar: How would we add the functionality to the camera manager? Would it check periodically if any respawn blocks are in the scene? I think the cleanest way is to have this functionality on the respawn block itself. I'm down to put it in some other script, though. Just didn't think it fit anywhere particularly well.

Assets/Scripts/Player/PlayerBounceBack.cs
AustinKellar: Sure, sounds good.

Assets/Scripts/BlockSpawner/Spawner.cs
AustinKellar: Yeah I agree. I think it would go in a separate PR, though.

AustinKellar: Though I do think that RespawnBlockInfo is best served as a private inner class. No other scripts should know it exists.

AustinKellar:
I like some of these suggestions. Which ones do you want in before merging?

Assets/Scripts/Blocks/CameraTracker.cs
IanGRap: I would say when the respawn block spawns it waits like 5 seconds and then adds itself to the camera manager's tracked transform. I was also wrong about this already being a function in camera manager. Maybe something like AddTrackedTransform after delay wold be a function worth making

AustinKellar: I don't think a time delay is a good idea, since the player can move the block while its off camera, which slows it down a lot. If we do it on a delay, we leave the possibility of the camera zooming out _a lot_.

AustinKellar: I honestly think the way we have it is the cleanest way.

IanGRap: okay that sounds like a good concern. Can we still move this code to a respawn block related function?

AustinKellar: Like rename the script to something with RespawnBlock in it?

IanGRap: I guess that would work. Maybe TrackTransformOnHeight? Camera tracker strikes me as too vague

AustinKellar: Agreed.

AustinKellar:
Alright, so I made some revisions.

1) Made a new script on the respawn blocks called PlayerRespawner that inherits from BlockDamager
    a) Took out the code from BlockDamager that was specific to respawn blocks
2) Renamed CameraTracker to RespawnBlockCameraController
3) Refactored collision events inside PlayerBounceBack
4) Put a skull inside the respawn blocks. 
    a) If someone could take a second pass at this, it would be much appreciated. I am not so good at making things look good in Unity.

Assets/Scripts/BlockSpawner/Spawner.cs
IanGRap: Maybe? I could see other scripts maybe needing to access this info in the future?

PR #154, opened 2019-04-17
Taunt button

AlfredoCuevas:
Contributors: Alfredo Cuevas

Goal: Remap the Y button to be the taunt / dab.

What Changed: 
* PlayerInput.cs: Now saves Y button press into the PressTaunt variable
* (New) PlayerTaunt.cs: Activates taunt and keeps a timer for the time taunting
* PlayerJump.cs: Added a check to prevent jumping while taunting
* PlayerMovement.cs: Added a check to stop movement/ prevent movement while taunting
* PlayerPrefab: I added the script to the prefab

How to Test: 
* Load any Scene with movable players. I used First Release. Also test in Lobby.
* Press Y to activate taunt
* Make sure there aren't any weird interactions with taunting by
  * Try moving around and taunting
  * Try jumping and taunting
  * Try dashing and taunting


Assets/Scripts/Player/PlayerTaunt.cs
```diff
@@ -0,0 +1,45 @@
...
+    void Awake()
+    {
+        _playerInput = GetComponent<PlayerInput>();
+        _playerDash = GetComponent<PlayerDash>();
+        _animator = GetComponentInChildren<Animator>();
+    }
+
+    void Update()
+    {
+        if (_playerInput.PressTaunt && !IsTaunting && !_playerDash.IsDashing)
```
IanGRap: I did notice that if you are in the air and press taunt you can't move left and right even though skelegoo won't start the taunt until he hits the ground. We'll have to see how playtesting goes, but we might want to disable taunting if you are jumping in the future.

JamesCoulter: I'm inclined to agree with this. If you are in the air and press Y you lose control and go straight down regardless of velocity, which doesn't feel good and could easily lead to accidental deaths.

PR #160, opened 2019-04-22
Issue 159 clearing problems

IanGRap:
Contributors: Ian

Goal: Fix problems where the lane clearer would not work when it needed to and sometimes it would clear blocks when it shouldn't have.

Important Files: 
- BlockClearer.cs: I made it so that the clear uses the block adjacency queries function instead of checking if velocity is zero. It turns out some blocks on the ground would have a velocity that wasn't zero, but would be so small that it wasn't noticable. I also removed the code that clears blocks on a timer because we haven't been using it and it was starting not to work with our implementation.
- Block Clearer.prefab: I reduced the height of the trigger box that keeps track of blocks on the floor. Turns out that it was too tall and including blocks that were one row above

How to test:
Load up sprint 5 scene and trying pushing blocks to fill the bottom row. It should clear whenever it fills and all the blocks have stopped moving. Also try pushing the long blocks down to see if they work to. I recommend setting the block spawner to primarily spawn long blocks.


Assets/Scripts/Environment Actions/BlockClearer.cs
```diff
@@ -51,77 +45,45 @@ private void OnTriggerEnter(Collider other)
...
+        _overlapBlocks.RemoveAll(obj => obj == null);
+
+        _overlapCount = 0;
+        foreach (GameObject obj in _overlapBlocks)
         {
-            return;
+            if (obj.tag == "LongBlock")
+            {
+                _overlapCount += 3;
+            } else
```
AlfredoCuevas: Needs to go on the next line.

PR #161, opened 2019-04-24
Squish detector fix

AustinKellar:
Contributors: Austin

Goal:
- Fix the duuuuuumb bug where players sometimes die when they aren't supposed to

What changed:
- Adjusted the position of the squish colliders to make them more responsive
- Changed the way bottomBlockIsNotFalling so that it uses the BlockAdjacencyQueries rather than the Rigidbody, because it is more reliable.

How to Test:
- Load up Sprint 5 scene and play for a little while. 
    - See if you get random deaths


AustinKellar:
Yeah there is. There is a check that the player isn't squished between the same object. Has to be 2 unique ones.

AustinKellar:
You know what, though? I've been using !=. I'm gonna try with !GameObject.ReferenceEquals()

IanGRap:
Still noticed bugs. Also, the options branch seems to be included here even though that branch hasn't been approved yet

PR #162, opened 2019-04-25
Camera revisions

IanGRap:
Contributors: Ian

Goal: Refactor and tweak the camera manager so that it better presents all that action to the players.

New Files:
- CameraTransformTracker.cs: Tracks all the objects that the camera is following. This code was originally in CameraManager.cs but it was getting cluttered so I moved it here. I also added a function that only tracks objects after they have fallen a certain distance. This function was the only code in RespawnBlockCameraController.cs That file is now deleted. The solo on player win variable tells the tracker if it should remove all transforms except the winning player when there is only one player left.

Important changes:
- BlockExploder.cs, Spawner.cs, PlayerRespawner.cs, MatchManager.cs, SceneCreator.cs: All these files now tell CameraTransformTracker.cs to track objects instead of CameraManager.cs
- CameraManager.cs: Added a camera zoom speed variable that lets you change how fast the field of view for the camera changes. Changed the map height and map width variables to a map size variable. The distances between the players are all compared to this value when deciding what the field of view of the camera should be. Increasing the horizontal buffer variable will decrease the influence the horizontal distance between players has on the field of view. For example, a horizontal buffer of 0 will mean that the FOV will be close to the max FOV when players are on opposite sides of the arena. Increasing the horizontal buffer will move the FOV towards the Minimum FOV in this case without affecting how the vertical distance between players changes the FOV. Also, when there is only one player left, the camera will zoom in dramatically and stop following other objects like bombs or respawn blocks.
- Sprint 5.unity: The UI manager had gotten dragged into the camera manager so I had to drag that out. Since I was already tweaking the scene, I turned off the white spotlights at the top of the arena because I thought they looked ugly and weren't helping the game.
- [CameraManger].prefab: I adjusted the values for the manager script so that the field of view would generally stay wider, hopefully making it easier to see when blocks were about to fall on you.

How to test: Play some matches on the Sprint 5 scene. The camera should focus on the winning player at the end of match. The camera should also better show what things are about to fall on the players and change FOV quick enough to keep players visible by the camera. What do you think?


Assets/Scripts/Other/CameraTransformTracker.cs
```diff
@@ -0,0 +1,119 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System.Linq;
+
+public class CameraTransformTracker : MonoBehaviour {
```
AustinKellar: Curly brace on the next line

PR #164, opened 2019-04-25
Win Screen Console Errors

AustinKellar:
Contributors: Austin

Goal:
- Get rid of the console errors that occasionally appear when the win screen starts up

What Changed:
- I changed the way the player -> color pair mapping is stored in ColorManager. It now has its own separate Dictionary data structure. 

How to Test:
- Play a bunch of games and see if you get a Null Reference Error when the win screen starts


AustinKellar:
Okay, after testing a while longer, I got it too. It's a dictionary error, meaning I asked for a key that wasn't inside the dictionary. I put an if statement around it: if (Dictionary.KeyExists(...), which seems to have solved the issue.

Assets/Scripts/Environment Actions/ColorManager.cs
```diff
@@ -83,19 +92,23 @@ private void SetNewPair(Scene scene, LoadSceneMode mode)
...
+            }
         }
     }
 
-    private string TrimObjectName(string name)
+    private void InitializePlayerColorMappings()
     {
-        if (name.Split('(').Length > 0)
+        _playerColorMappings = new Dictionary<Color, ColorPair>();
+        for (int i = 0; i < _playerMaterials.Count; i++)
```
IanGRap: we should throw an error if the playerMaterials count does not equal the correspondingPlayerColorPairs count.

AustinKellar: will do

PR #166, opened 2019-04-26
Stopping player collisions

IanGRap:
Contributors: Ian

Goal: Add functionality for players to be able to move through each other to stop body blocking.

Important Files:
- PlayerDash.cs: Setting Ignore Player Collisions to true means that players will never collide with other players. Setting Ignore Player Collision On Dash to true means player collisions will only be ignored if at least one player is dashing. These settings get set globaly, meaning that if the options are set to true on one skelegoo, it will effect all skelegoos. I recommend only changing this setting on the prefab itself.
- Lobby.unity: The skelegoo on the title screen was falling through the floor so I had to change the collision layers of the blocks below him.

How to test: Load up the sprint 5 scene and play around. You should pass through all other skelegoos always. Go to the prefab and uncheck Ignore Player Collisions then check Ignore Player Collision On Dash. When you play the scene again you should only be able to pass through players when you dash. Uncheck both options and you will not be able to pass through players at all. Right now the default is always pass through players. What do you think?


IanGRap:
Looks like the prefab values were messed up, should be good now

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -36,6 +36,15 @@ public class PlayerDash : MonoBehaviour
     [SerializeField]
     private bool _normalizeDiagonalDash = true;
 
+    [SerializeField]
+    private bool _ignoreCollisionsWithPlayers;
+
+    [SerializeField]
+    private bool _ignorePlayerCollisionsOnDash;
```
AustinKellar: Can we throw an error if both of these values are true? From my understanding, you would only want one or the other to be true at any given time.

AustinKellar:
When you have Ignore Player Collision On Dash set to true, then you dash _through_ a player and into a block, it doesn't push the block. It feels super weird to me. I don't think its really a big deal though, since we will probably have Ignore Collisions With Players be the default setting.

PR #169, opened 2019-04-27
Ui refactor

AustinKellar:
Contributors: Austin

Goals:
- Refactor the UI code so that shared logic is all contained to one file
- Reorganize the file structure with respect to the UI scripts
- Implement an option where each menu can use the ColorManger's color as the button colors

What Changed;
- UIController
    - The new file that contains the shared menu logic
- IMenuInput
    - An interface for menu input capturing scripts
- UIMatchMenuInput
    - Was UIMenuInput, renamed to UIMatchMenuInput, implements IMenuInput
- UILobbyMenuInput
    - Now implements IMenuInput Interface
- UITitleMenu, UIOptionsMenu, UIPauseMenu, UIScoreMenu, UIWinScreenMenu
    - All of these have the shared logic removed and inherit from UIController
- All UI code was moved into a new folder called Assets/Scripts/UI

How to Test:
- Make sure all menus still work as expected.
- Make sure they also work in a build
- Make sure you can use ColorManager colors by changing Use Color Manager Colors on each menu
- Do you like the code changes? Any suggestions?
- Do you like the new folder structure?


Assets/Scripts/UI/Match/UIScoreMenu.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        base.menuInput = UIMatchManager.Instance.GetComponent<UIMatchMenuInput>();
+
+        base.Initialize();
+    }
+
+    protected override void OnMenuStart()
+    {
+        SceneCreator sceneCreator = MatchManager.Instance.GetComponent<SceneCreator>();
+        sceneCreator.AllocatePlayers();
```
IanGRap: I don't feel like this UI manager should be telling the scene creator when to set up the scene (It feels outside the domain of what a UI controller would be expected to do). Can this be done elsewhere? Maybe just on the start function of the Match manager

Assets/Scripts/UI/IMenuInput.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public interface IMenuInput
```
IanGRap: Should this be called UIMenuInput?

AustinKellar: No, the C# naming convention for interfaces is to prepend an I. So it could be IUIMenuInput but I thought that was too weird

Assets/Scripts/UI/Match/UIScoreMenu.cs
AustinKellar: Yeah, sounds good. I didn't notice this was here, honestly. Just copied it from the start function in the file

AustinKellar:
I did have the colors on the win screen menu originally but I thought it looked weird with some skelegoos. Could you try enabling it and letting me know what you think? (If you have time)

JamesCoulter:
Also, the issue where players can change their selection during the win screen after an option has already been picked is still present, and should probably be removed. Dunno if that should be separate from this pr

AustinKellar:
Thanks, James. I solved both issues that you pointed out and made the win screen use the ColorManager colors as Ian suggested. Could you give it another look?

PR #171, opened 2019-05-02
Better cross fade

AlfredoCuevas:
Contributors: Alfredo Cuevas, (Fernando for last 3 commits)

Goal: Make a system that allows for multiple songs to be played as if they were one song. Have the songs CrossFade when they change. Make sure that all the functionality goes through the AudioManager so that audio functions stay consolidated in one place.

What Changed: 
* **AudioManager.cs**: Added new functions that allow programmers to play, stop, fadeIn, and fadeOut the gameplay music.
  * IsGameplayMusicPlaying(): returns true if gameplay music is playing, false otherwise
  * PlayGameplayTheme(): starts GameplayTheme. Has two parameters that allow for choosing the first song
  * StopGameplayTheme(): stops all GameplayTheme music
  * FadeInGameplayTheme(): basically the same as play but lets you choose a longer fade time
  * FadeOutGameplayTheme(): basically the same as stop but lets you choose a longer fade time
* **DynamicGameplayTheme.cs**: This class keeps track of all the important aspects of the Gameplay Theme. It is made to store lists of songs and randomly cycle between the songs to create a seamless sound experience. AudioSources that are used in the Gameplay theme should not be controlled through regular AudioManager functions. 
* **GameplayThemeTester.cs**: A Temporary script that listens to key inputs. Used for testing

How To Test: 
* Load into the BetterCrossFade scene
* Play the scene. At first nothing will happen.
* Press Up Arrow to call PlayGamePlayTheme(). Press Down Arrow to call StopGamePlayTheme
* Press W to call FadeInGameplayTheme(). Press S to call FadeOutGameplayTheme()
  * You can change the fade time value by changing the numbers on the GamePlayThemeTester GameObject. Don't make them negative
* Play around with these keys. See if any glaring bugs pop up. 
* Let the music play for a few minutes. Make sure that the crossfade between multiple songs works
* During a Crossfade try to stop/fadeOut the gameplay theme. Does it work as expected? 
* Enjoy Fernando's Music because you can only test s crossfade by waiting until the song ends
* You can tell when the CrossFade is happening by opening the AudioManager during play and selecting one of the GT_SongName objects. Look at the volume and wait until it starts fading. 

**Notes**: I have begun using the naming convention of GT_SongName for gameobjects in the AudioManager that are used for Gameplay Theme purposes. These songs should **not** be called through regular AudioManager functions like Play()/PlayOneShot().

I think more songs are needed to make this new Feature worth using

I didn't get around to adding any list order functionality. I wanted to see if people liked this before delving deeper. 

------------------------------------------------------------------------------------------------------------------
Last 3 commits now include some updates to the AudioManager and AudioMixer. This was done in collaboration with Fernando.


Assets/Scripts/DynamicGameplayTheme.cs
```diff
@@ -0,0 +1,236 @@
...
+{
+    [SerializeField]
+    private AudioSource[] _audioPieces;
+
+    [SerializeField]
+    private AudioSource[] _audioPieces2;
+
+    // private fields-----------------------------------------------------
+    private const float _songFadeWindow = 30f;
+    private AudioSource[][] _audioLists;
```
IanGRap: I'm confused on what the reasoning for a 2D array for storing the songs is? Why can't we just do a list?

AlfredoCuevas: This is for future song control. Right now it wasn't really something that I could test. I need more songs. In the future you should be able to split different types of songs into categories. Then the Queue will pick from a specific list based on an order given. Right now the multiple lists don't do anything. 
Fernando mentioned that he may want this feature so I wrote the script with this in mind.

IanGRap: ahhh for sure. What works for Fernando works for me

IanGRap: Is there a way to just have the songs loop forever?

AlfredoCuevas: I'm not sure what you mean. Right now the song will loop. If you are talking about looping an individual song then there is no point in using this script. Simply call play through the AudioManager.Play().

AlfredoCuevas:
> Code looks good and it sounds great. I feel a little weird about merging in your test file though, should we maybe integrate with gameplay before merging? Also, please remove debug.logs before merging.

I'll delete the test script but for now I don't want to fully integrate to avoid any conflicts. Ill integrate the function calls in my next pull request once we have more finalized music.

PR #174, opened 2019-05-03
Respawn tweaks 1

JamesCoulter:
Contributors: Coulter, Ian

Goal: Make various tweaks to the way respawn blocks work, with the high-level goal being to make them more powerful and more impactful during matches. Also Adding placeholder SFX and UI indicator.

Files Changed:
- Spawner.cs: Added a respawn function specifically for respawn blocks.
- BlockInput.cs: Repurposed old script to allow players to move their respawn blocks using thumbstick inputs.
- RespawnUI.cs: New script that spawns the UI image for each respawn block and destroys the image when the block appears on the screen.
- PlayerRespawner.cs: New script that respawns the player when they kill someone as a block.
- Respawn Canvas: New canvas that spawns and holds the UI images as its children.
- Respawn Image: A skelegoo icon + an arrow pointing upward, spawned by the Canvas with the color of the corresponding respawn block.
- BlockCollisionDelegator.cs: Fixed a bug that allowed players to push long and tall blocks upward.
- Respawn Beam: Made the beam a child of the respawning player on instantiate(), moved the transform to avoid breaking blocks below the respawning player.

How to test: 
- Have a player die. Their respawn block should spawn in immediately, with a UI indicator at the top of the screen that tracks them until they enter the camera view.
- Move the block using the left thumbstick. It should be able to move and push other blocks in all four directions, but cannot break itself against the environment. Can be pushed and broken by players. It should also no longer break once its velocity is zero (i.e. reaching the ground or landing on a stationary block)
- Use the respawn block to kill someone. The controlling player should instantly respawn at the location of the kill, with a beam that destroys all blocks directly above the respawned player.

Notes: The code for this one is probably in need of some refactoring and triage, so any suggestions for improving it are welcome. Ideas for improving the UI are also welcome. Shout-out to Ian for helping me work through getting the UI set up. Also, there is a known bug that I discovered and decided to leave out of this PR, wherein two players who die simultaneously will have their respawn blocks spawn inside one another.


Assets/Scripts/Blocks/BlockInput.cs
```diff
@@ -29,13 +33,39 @@ private void Update()
             return;
         }
         _gamePadState = GamePad.GetState(Index);
+        if (_gamePadState.GetAxis(CAxis.LY) < 0f && _movementYAxisLastFrame >= 0f)
+        {
+            _blockPusher.Push(Vector3.up);
+            _lastPushTime = Time.time;
+        }
```
IanGRap: theoretically, a player could push themselves up repeatedly and then leave the arena. Fun thought

Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -78,15 +78,15 @@ private IEnumerator SquashAndStretchDelay(float delay)
     public virtual bool Push(Vector3 direction, float animationDelay = 0f)
     {
         // TEMPORARY to test not being able to push up
-        if (IsPushed || direction == Vector3.up)
+        if (IsPushed && gameObject.tag == "Block" || direction == Vector3.up && gameObject.tag == "Block")
```
IanGRap: Can we change this now that the BlockCollisionDelegator stops players from pushing up?

Assets/Scripts/Blocks/BlockSpawner/Spawner.cs
```diff
@@ -117,6 +126,18 @@ private void Start()
         StartCoroutine(SpawnTimer());
     }
 
+    private void Update()
+    {
+        _playerToRespawn = _playersToRespawn.OrderBy(p => p.RespawnQueueTime).FirstOrDefault();
+        if (_instantRespawn)
+        {
+            if(_playerToRespawn != null)
```
IanGRap: maybe turn these 2 if statements into 1 with a &&

Assets/Scripts/Blocks/RespawnPointController.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    private void OnTriggerEnter(Collider other)
+    {
+        _objectsInRespawnArea.Add(other.gameObject);
+    }
+
+    private void OnTriggerExit(Collider other)
+    {
+        _objectsInRespawnArea.Remove(other.gameObject);
+    }
```
IanGRap: This all seems really similar to the code in Spawner.cs. Maybe this tracking should be made into its own class? Later PR thing

Assets/Scripts/Match UI/RespawnImageController.cs
```diff
@@ -0,0 +1,16 @@
...
+	// Use this for initialization
+	void Start () {
+		
+	}
+	
+	// Update is called once per frame
+	void Update () {
+		
+	}
+}
```
IanGRap: can we delete this file?

Assets/Scripts/Match UI/RespawnUI.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    private void Start()
+    {
+        _cam = CameraManager.Instance.GetComponent<Camera>();
+        _imagesWithTransforms = new Dictionary<Transform, GameObject>();
+    }
+
+    private void Update()
+    {
+        _count = _imagesWithTransforms.Count;
```
IanGRap: we can get rid of count because that was for testing right?

Assets/Scripts/Match UI/RespawnUI.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    [SerializeField]
+    private GameObject _respawnImage;
+    private float _max;
+    private int _count;
+    private Spawner _spawner;
+    private Camera _cam;
+    private Transform _blockTransform;
+    private Vector3 _screenPos;
+    private Vector2 _onScreenPos;
```
IanGRap: why do these vectors need to be declared here. Aren't they only used in 1 function?

Assets/Scripts/Match UI/RespawnUI.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (_screenPos.x >= 0 && _screenPos.x <= 1 && _screenPos.y >= 0 && _screenPos.y <= 1)
+            {
+                imagesToDestroy.Add(pair.Key);
+                continue;
+            }
+
+            _onScreenPos = new Vector2(_screenPos.x - 0.5f, _screenPos.y - 0.5f) * 2; //2D version, new mapping
+            _max = Mathf.Max(Mathf.Abs(_onScreenPos.x), Mathf.Abs(_onScreenPos.y)); //get largest offset
+            _onScreenPos = (_onScreenPos / (_max * 2)) + new Vector2(0.5f, 0.5f); //undo mapping
+            _onScreenPos = _cam.ViewportToScreenPoint(_onScreenPos);
```
IanGRap: what was the reasoning behind all this math again?

Assets/Scripts/Blocks/BlockPusher.cs
IanGRap: In playtesting pushing single blocks up didn't seem to be working right

Assets/Scripts/Match UI/RespawnImageController.cs
JamesCoulter: Thought I had deleted this one already, my bad.

Assets/Scripts/Match UI/RespawnUI.cs
JamesCoulter: _cam is initialized on start(), but you're right for the other ones.

Assets/Scripts/Match UI/RespawnUI.cs
```diff
@@ -0,0 +1,91 @@
...
+            return;
+        }
+        GameObject imagetoDestroy = _imagesWithTransforms[t];      
+        _imagesWithTransforms.Remove(t);
+        Destroy(imagetoDestroy);
+    }
+
+    private void UpdateUIPositions()
+    {
+        List<Transform> imagesToDestroy = new List<Transform>();
```
AustinKellar: I think we should make a private field out of imagesToDestroy. It is really expensive to create a new list every frame. It could really hike up the garbage collector. Instead of creating a new list every frame, could we create it once and add/remove from it as needed?

AustinKellar:
I'm concerned about this for a few reasons. The UI looks great and I think its an awesome change, but the rules for when respawn blocks get destroyed are extremely unclear to me. While I was playing, I occasionally got blown up as a respawn block and I could not figure out the pattern. Also, if you kill someone by chain pushing blocks, you do not respawn, which feels cheap to me. I think either you should respawn by chain killing or _not_ be able to push blocks. I'm also quite worried that the respawn blocks will be OP now. At the very least, they need some more hard rules for when they get destroyed. Maybe a timer? Maybe when they hit the ground? Maybe when a player dashes into it? Right now they really don't feel good to me. But I didn't see any functionality issues as you described in the PR, so I'll approve.

JamesCoulter:
Like I say above, part of the goal for the changes in this PR was to make the blocks more powerful than they were initially. It's a big feature in general, so I imagine this is just the first round of tweaks to make it feel the way we want. Lots of playtesting is going to be needed. Your comment about clearer breakage rules is one I agree with, though.

IanGRap:
Austin I would like to see playtesting show that these additions are too powerful before reducing them. I still want to make sure this feature is fun before focusing in on balance. Do you have any more information on when the block broke randomly. The only one that seems like it might be confusing to me is that the respawn blocks all get destroyed at the end of the round. All the other ones seem pretty straight forward to me. We also have a story for the next sprint to do another update on the block

AustinKellar:
Its not that the block broke randomly, more that through playing for 10 minutes, I could not figure out the cause of it breaking. It _felt_ random. Idk, just didn't quite feel good. I think down the line we should have a different set of break conditions.

IanGRap:
Ahh I see. One of the stories for the next sprint is to do another revision of the goo/respawn block with a bigger focus on design. I think we will have a playtest and discussion at our next design meeting to see where to go from here

PR #175, opened 2019-05-06
Sprint 6 build

IanGRap:
Contributors: Ian, Liam

Goal: Get the latest build updated with all art assets and features. Also make a few tweaks along the way.

Code Changes:
- BlockDamager.cs: Added flag to let you choose if the break particle should have the block material when it spawns. Fixes problem where death and bomb blocks spawned pink particles
- BlockSpawnLocationTracker.cs: Now chooses a random player to start spawning blocks over instead of spawning on top of just the first player.
-DJAnimationDirector.cs: Tells the DJ to switch animations on a timer.
- ColorManager.cs: Fixed a mistake with event listening.
- CameraManager.cs: Made it so camera shake can play even if there aren't transforms being tracked so the camera will shake when the game starts.
- WinScreenController.cs: Added a function to delay when to play the win sound so it syncs up.
- SkinnedColorSelector.cs: Basically the same as the color selector but now it works for skinned meshes. Also, you can choose the intensity of the emissive color.
- Wall.cs: Made it so you can have back and forth colors for multiple emoji walls. The multiple emoji walls were not popular with the artists so I took it out but left the code in if we ever want to go back.
- UIScoreImage.cs: Made it so the displayed score is out of the score needed to win.
-Spawner.cs: Added an additional check before spawning a block to avoid respawn blocks spawning at the same time as other blocks.

Other Changes:
- Put in retopologized backgrounds.
- Added chonkier particles to block breaks.
- Added an animated DJ in the background.
- Synced up crossfader and multiple songs with gameplay.
- Reduced the camera zoom when someone wins
- Reduced the vignette in post processing and added anti aliasing.
- Added new rubble art asset
- Added new pillars.
- Added moving spotlights.
- Removed black color combos
- Removed UI colors on title screen, win screen and score screen
- Added a red volumetric light on the death skull to make it more visible and give players more warning
- Added the win sound on the win screen
- Show the win score on the score screen
- Added a dance block behind the compacting walls to obstruct things that are not in the game any more
- Swapped candles out for low poly versions
- Turned off most of the lights in the scene to help performance

How to test: 
Play through the lobby scene to gameplay to win a few times. Test in build. Should any small changes be made? Should any changes be rolled back?


Assets/Scripts/DJAnimationDirector.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
```
AlfredoCuevas: You could probably include a [Requires(typeof(Animator))] here

Assets/Scripts/DJAnimationDirector.cs
```diff
@@ -0,0 +1,29 @@
...
+public class DJAnimationDirector : MonoBehaviour {
```
AlfredoCuevas: The { should be on the next line.

Assets/Scripts/Environment Actions/SkinnedColorSelector.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using DG.Tweening;
+
+public class SkinnedColorSelector : MonoBehaviour {
```
AlfredoCuevas: { needs to be on the next line

PR #183, opened 2019-05-08
Rewired

AustinKellar:
Goals:
- Switch input system to Rewired
- Added controller vibration in key areas

What changed:
- Installed Rewired
- Created a wrapper class around Rewired called CInput. Its a static class that models CarbonInput
- Adapted all input reading files so that they use the new system
- Added vibration calls throughout the project

How to Test:
- Play the game and make sure all of the controls still work as expected.
- Make sure you feel the controller vibrate in the following scenarios:
    - Lobby
        - When you join
        - When you leave
        - When you quit to title menu
        - When you exit the "Begin Rave" scene
        - When the floor breaks
    - Match
        - When you die
        - When lane clears
        - When bomb block goes off
    - General
        - When blocks break in camera transition
- Does the wrapper class feel clean?
- Does the vibration feel good?


Assets/Scripts/Player/PlayerInput.cs
```diff
@@ -123,23 +119,14 @@ private void CaptureHorizontalMovementInput()
...
-            _gamePadState.Pressed(CButton.X) ||
-            _gamePadState.Pressed(CButton.B) ||
-            _gamePadState.Pressed(CButton.LB) ||
-            _gamePadState.Pressed(CButton.RB) || 
-            _gamePadState.Pressed(CAxis.RT) ||
-            _gamePadState.Pressed(CAxis.LT) ||
-            _gamePadState.Pressed(CButton.LS) ||
-            _gamePadState.Pressed(CButton.RS)
-        );
+        PressDash = CInput.GetButtonDown(Index, CButton.X);
```
IanGRap: My personal thought is we should still have the triggers and bumpers as options for dash (not b though). I imagine rewired will probably be better at handling triggers than carbon. Also, I noticed some of my controllers just don't have working triggers or bumpers so that might have influenced us on the triggers not working

Assets/Scripts/Public Helper Classes/CInput.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using Rewired;
+using System.Linq;
+
+public static class CInput
+{
+    public static bool GetButton(PlayerIndex index, CButton b)
```
IanGRap: can we change b to button? b makes me think the b button

Assets/Scripts/Public Helper Classes/CInput.cs
```diff
@@ -0,0 +1,86 @@
...
+        else
+        {
+            ReInput.players.GetPlayer(index.ToString()).SetVibration(0, intensity, duration);
+        }
+    }
+}
+
+public enum PlayerIndex { Any, One, Two, Three, Four, All, None }
+public enum CButton { A, B, X, Y, Start, Back, Up, Down, Right, Left, LB, RB, LS, RS }
+public enum CAxis { LX, LY, RX, RY, LT, RT }
```
IanGRap: This code looks good. The wrapper class was a really good idea!

Assets/Scripts/Public Helper Classes/CInput.cs
AlfredoCuevas: ^

Assets/Scripts/Player/PlayerInput.cs
AlfredoCuevas: I agree that we should include triggers and bumpers IF they actually work consistently

PR #184, opened 2019-05-08
Audio manager updates

AlfredoCuevas:
Contributors: Alfredo

Goal: Add a new function that will choose from an array of Block Break Sounds (provided by Fernando) and pitch shift them slightly every time they are called. 

Changes: 
* AudioManager.cs: Updated the class to use a Dictionary when searching for songs. Added a new function that calls VariableSoundPitchShifter.cs
* VariableSoundPitchShifter.cs: This takes a list of block break sounds. It has a public function that will PlayOneShot() a random block break sound and pitch shift it slightly. 
* CameraManager.cs, PlayerSelectBlockSpawner.cs: these have been changed to call the new PlayOneShotBlockBreak()
* BlockDamager: This used to call Play("") but it has been changed to PlayOneShotBlockBreak()

How to test: 
* Play through all parts of the game. Make sure that the block breaking sounds are ok. Try to listen for the slight changes in the block breaking sounds. 
* After you have done this Play the game but this time go into the AudioManager and in VariableSoundPitchShifter change the range of Min and Max. You should notice a more drastic pitch shift. 1== normal pitch.
----
Implemented a suggested change to make a class to hold the info needed for any new sound groups


Assets/Scripts/VariableSoundPitchShifter.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+[RequireComponent(typeof(AudioManager))]
+public class VariableSoundPitchShifter : MonoBehaviour
+{
+    [SerializeField]
+    private AudioSource[] _blockBreakSounds;
```
IanGRap: It does seem weird to me that this is hard coded to only support the block break sounds. It might be worth making an internal serializable class called audio groups. Each one has a group name (example: "Block Breaks"), a list/array of the possible sound variations and a min pitch and max pitch. Then you just have a list of audio groups. If the class is serializable, then it can be edited from the editor. However, if we are only going to need this variation for the block break sounds and nothing else, then maybe this isn't worth the investment

AlfredoCuevas: I think this is a good idea. I can't believe I didn't think about making each group a class. For some reason my brain wanted to make it a Scriptable Object. I think that Fernando also wanted dashing to be have multiple sounds. I'm not sure if there is anything else though.

IanGRap: Sounds like we should do it

PR #185, opened 2019-05-09
Tweaking block spawn logic

IanGRap:
Contributors: Ian

Goal: Make it so the block spawner only spawns hazardous blocks when there are a lot of blocks in the scene and more chaotic block formations (like ones that include multiple long or tall blocks) spawn later in the match.

Small Changes:
-BlockDamager.cs: When a block spawn it tells the spawner to increase the block scene count based on a value. A normal block has a value of 1, where as a long or tall block has a value of 3 (its basically 3 blocks in the scene). Hazardous blocks like the death block have a value of 0 since they remove blocks from the game.
- BlockFormationData.cs: Removed the _lowestBlockWidth variable since it was no longer being used.
- WallManger.cs: Added an event listener that sends actiosn whenever the walls compact.

Spawner.cs Changes:
- When blocks are spawned or destroyed they can have different values toward the block count (outlined in the BlockDamager.cs changes).
- Added a private class called block wrappers. This contains the tags, prefabs and hazard delay(more on this later) for the different types of blocks. In the Spawn function, instead of creating an if else statement for every type of block, the tag of the block in the formation is compared to what wrapper it matches and that block is spawned.
- Now a list for hazardous blocks. When choosing a transform, the closer the block count is to high block count, the higher the likeliness of spawning a hazard. If the hazard has a cooldown set in the wrapper, another hazard will not be spawned until the cooldown has expired.
- Refactored the spawner so that it doesn't create the template prefab of a block formation when it spawns. It instead just references the prefab and places blocks in the corresponding world positions. these blocks don't have a parent any more, so the temporarily track transform would no longer work and was removed (we weren't using it anyway).
- The block count no longer influences the spawn rate of blocks. It now just influences the likeliness of hazards spawning.
- Created a private class called delayed block formations. It tracks block formations with a required number of times the walls need to compact before it can be used. There is a list of delayed block formations that are only added to list of possible formations after the required number of wall compacts.
- The OnWallsCompact function listens to the WallManager event. It reduces the number of the high block count to influence hazards so hazards still spawn when the arena gets small. It also adds new block formations.

How to test:
Load up the sprint scene. Hazards (death block and bomb block) generally spawn more frequently when the arena fills with blocks. Hazards should not spawn right after one another. The block formations that spawn initially should only include normal blocks. As the walls compact more, you should see an increase of formations that include the long and tall blocks spawning. On the spawner, you can adjust the cooldown of hazards by changing the hazard cooldown variable on the wrapper list. You can change which formations become available and when by going to the delayed formations list and changing the required compact count field.

Notes:
I didn't have a group to playtest with so a lot of these values will need to be tweaked. This is more so a proof of concept.


Assets/Scripts/Blocks/BlockDamager.cs
```diff
@@ -16,6 +16,8 @@ public class BlockDamager : MonoBehaviour
     private List<Texture2D> _damageTextures;
     [SerializeField]
     private bool _setColorOfBreakParticleOnBreak = true;
+    [SerializeField]
+    private int _blockValue = 1;
```
AustinKellar: This variable name is a little vague to me. What is blockValue? Since its a variable in the BlockDamager I think maybe it's the number of times it has been pushed?

Assets/Scripts/Blocks/BlockSpawner/Spawner.cs
```diff
@@ -33,24 +34,34 @@ public class Spawner : MonoBehaviour
...
     [SerializeField]
-    private bool _instantRespawn = false;
+    private List<Transform> _shapePrefabs;
 
     [SerializeField]
-    private Transform[] _shapePrefabs;
+    private List<DelayedBlockFormation> _delayedBlockFormations;
+
+    [SerializeField]
+    private List<Transform> _hazardPrefabs;
```
AustinKellar: Since these are block types, could we rename this to something like _hazardBlockPrefabs? _hazardPrefabs makes me think of the wall movement as well.

PR #186, opened 2019-05-09
Dash cooldown feedback

JamesCoulter:
Contributors: Coulter

Goal: Add visible feedback that indicates when the dash cooldown ends for a player after dashing.

Files Changed:
- PlayerDash.cs: Within the dash() function, the player's material color is darkened and their point light is disabled after dashing. They remain this way for the duration of the cooldown, and then return to normal to indicate that a player can now dash again. An OriginalColor public property has been added so that the default colors for each player material can be set and stored.
- PlayerColorController.cs: Added lines to set the correct original and cooldown colors in the PlayerDash script. Works only for Sprint 6 scene.
- PlayerSelectColorController.cs: Added lines to set the correct original and cooldown colors in the PlayerDash script. Works only for the lobby scene.
- PlayerRespawner.cs: Added lines to set the correct original and cooldown colors in the PlayerDash script for respawned players.
- SquishDetector.cs: Added a line that resets a player's material right before they die, so that the correct material is passed to the camera splat, respawn block and UI indicator.

How to test: 
- Start a game in either the lobby or Sprint 6 scene.
- Dash. The player point light should disappear and their color should darken.
- Observe whether or not the point light and color are correctly reset once the dash cooldown ends. Both should just return to their original states.
- Attempt to dash right before you get squished, so that your color is darker when you die. The goo splat, respawn UI, and respawn block color should all be the original color.
- Do the same thing again, and successfully respawn. The respawned player color should be correct.

Notes:
It seems that Unity does not like it when you directly change the color of a material. Doing so causes new instances of the material to be made each time. On top of being the source of the major bug with the respawn colors, this also takes up unnecessary memory. The better approaches are to either change the color through shaders, or have separate materials for each color. I'm now using the second approach, wherein two materials are set in the PlayerColorController and PlayerSelectColorController scripts, and these two materials are swapped out in the skinnedmeshrenderer component as needed. It's ugly, but it's the only way I've been able to make it work.


Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -52,12 +62,22 @@ public class PlayerDash : MonoBehaviour
     private PlayerMovement _playerMovement;
     private PlayerGrounded _playerGrounded;
     private Animator _animator;
+    private Light _playerLight;
+    private SkinnedMeshRenderer _playerRenderer;
+    private Material _playerMat;
 
+    private float t;
```
AlfredoCuevas: Not used in the script.

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -16,6 +17,15 @@ public class PlayerDash : MonoBehaviour
...
     private float _dashCooldown;
 
+    [SerializeField]
+    private float _dashCooldownPercent = 2.5f;
+
+    [SerializeField]
+    private float _cooldownFadeOut = 0.5f;
+
+    [SerializeField]
+    private float _cooldownFadeIn = 0.01f;
```
AlfredoCuevas: These three variables don't seem to be used in the script.

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections;
+﻿using DG.Tweening;
```
AlfredoCuevas: Not used in the script

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -52,12 +62,22 @@ public class PlayerDash : MonoBehaviour
...
     private float _previousDashTime;
     private bool _groundedSinceLastDash;
     private float _tolerance;
     private int _playerLayer;
 
+    public Material OriginalMat { get; set; }
+
+    public Material CooldownMat { get; set; }
+
+    public float DashCooldownPercent { get; private set; }
```
AlfredoCuevas: Not used in the script

JamesCoulter:
Agh, so many variables no longer being used that I forgot to remove. This is what I get for writing code after getting very little sleep

Assets/Scripts/Environment Actions/Managers/PlayerColorController.cs
```diff
@@ -5,6 +5,9 @@
 
 public class PlayerColorController : MonoBehaviour
 {
+    [SerializeField]
+    float _cooldownPercent = 2.5f;
+
```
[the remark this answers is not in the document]
JamesCoulter: They shouldn't affect each other since the process for getting the cooldown color is separate for each scene, but it also just looks bad. I'm gonna do another pass

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -130,6 +144,11 @@ private void FixedUpdate()
 
     private void Dash()
     {
+        _newColor = new Color(OriginalColor.r - _cooldownDarken,
```
IanGRap: why do we need to remake this color everytime we dash. Can't we just make the color on awake once?

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -52,11 +58,17 @@ public class PlayerDash : MonoBehaviour
...
     private Animator _animator;
+    private Light _playerLight;
+    private SkinnedMeshRenderer _playerRenderer;
+    private Material _playerMat;
 
     private float _previousDashTime;
     private bool _groundedSinceLastDash;
     private float _tolerance;
     private int _playerLayer;
+    private Color _newColor;
```
IanGRap: new color as a variable name seems slightly confusing. Can it be _dashCooldownColor?

JamesCoulter:
@IanGRap The OriginalColor variable in PlayerDash.cs needs to have its value set every time a player object has its color changed, meaning in the lobby, at the start of the main scene, and every time a player respawns in the main scene. If it is set anywhere other than at each of these points, the player color after dashing is wrong. That's why PlayerSelectColorController.cs, PlayerColorController.cs, and PlayerRespawner.cs all set the value. 

OriginalColor can't be set on Awake() because individual player colors in the main scene aren't set until Start() in SceneCreator.cs (where AssignPlayerColors() is called). If I set it on Awake(), it will always be gooey green.

PR #188, opened 2019-05-10
Push with button press

IanGRap:
Contributors: Ian

Goal: Address issue 176 and make it so you dash in the direction you are facing when you press a dash button but don't specify a direction.

Changed Files: 
- PlayerRotator.cs: You can now access which direction the player is facing through a property. I also added an enum that specifies the direction the player is currently facing to make it more readable. There is a stun function that temporarily turns of direction changing. This is so the player doesn't immediately turn around after bouncing back from a dash. This way you can keep pushing in the same direction with one button press.
- Added additional checks to dash function where if there is no direction specified by the joystick, the dash direction defaults to a horizontal dash in the direction you are facing.

How to test: Load up the lobby or match scene and try dashing without pressing a direction. You should dash in the direction skelegoo is facing. You should still dash in the direction you indicated with the joystick when you hold it in a direction. This should also work when you set player dash to use the exact direction of the joystick press. When you dash and hit a box, then dash again by only pressing the push button, you should dash in the same direction.


Assets/Scripts/Player/PlayerRotator.cs
```diff
@@ -7,10 +7,15 @@ public class PlayerRotator : MonoBehaviour
     [SerializeField]
     private Transform model;
 
-    private bool _facingRight = false;
+    private bool _facingLeft = false;
```
AustinKellar: In a situation like this, I think its better to use an enum than a boolean. There could be a private enum called Direction { Left, Right }. To me, its cleaner to read than using a bool but its up to you.

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -162,7 +164,14 @@ private void Dash()
 
             if (_exactDirection)
```
AustinKellar: Did you handle the case if _exactDirection is off? I know we haven't used it since the beginning of the project but for consistency's sake, we may want to support it.

IanGRap: yeah it should work. Did you test it?

PR #190, opened 2019-05-12
Mixer snapshot transitions

AlfredoCuevas:
Contributors: Alfredo, Fernando

Goal: Add Audio Mixer snapshots to the game so that we can have different mix levels while in the lobby, gameplay, and pause states. 

What Changed: 
* Audio Mixer: Got the new Pause snapshot mixed by Fernando.
* MixerSnapshot.cs: holds the info for the different snapshots. Has a function called by the AudioManager that will transition to a given snapshot. 
* AudioManager.cs: Calls the function in MixerSnapshots.cs. Now holds an enum for the different available snapshots. These are used when calling the new function
* UIMatchManager.cs: Makes the transitoin calls for pausing the game and going back to the lobby from the pause menu
* WinScreenController.cs: makes the call to transition snapshots from the win screen to the lobby. If you restart a match it will simply stay in the gameplay snapshot mode. 
* PlayerSelect.cs: Transitions into the gameplay snapshot. 
* I also added Fernando's new looping main menu theme and updated the AudioManager prefab

How to Test: 
* Load into the Lobby Scene. Make sure you have the Audio Mixer tab open
* While you are listening try to keep an ear out for any thing that sounds too abrupt or not right.
* Start playing. Notice in the Snapshots section of the mixer it should say you are on the Lobby preset
* Transition into the game. When you transition keep an eye on the snapshots. It should switch to Gameplay
* Pause the game and Unpause it. The snapshots should switch between Pause and Gameplay. This transition should be the most noticeable. 
* Go from the Pause menu back to the lobby. Make sure that transition happens successfully. 
* Then finish a game and Go from the Win screen back to the lobby. The transition in the mixer should go from Gameplay -> Lobby

**Note**: Currently the transitions are made to start in the Lobby. If you start a game from the Sprint6 scene it will start you in the Lobby snapshot preset. The snapshots do not read the state of the game. (I don't think we currently hold that info anywhere). If you NEED to be in the gameplay snapshot and start from sprint 6 then you can just Pause and unpause the game. This will cause a transition from Pause -> Gameplay.


Assets/Scripts/Audio/MixerSnapshots.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Audio;
+
+public class MixerSnapshots : MonoBehaviour
```
IanGRap: Minor but this class name seems weird to me. MixerSnapshots makes me think it is just the snapshots. Maybe something like snapshot controller?

AlfredoCuevas: I'm not sure about calling it a controller though because it only has one function that switches snapshots when called. In my mind this class is just a container for the snapshots. The one function is only there to make sure the AudioManager doesn't get cluttered.

PR #191, opened 2019-05-13
Color blindness

IanGRap:
Contributors: Ian, people that were in the room

Goal: Make character colors more accommodating to those with color blindness.

Color Changes:
- Oozing Orange is now much more peach color
- Yucky Yellow has been removed
- Rotten Red has been removed
- Malicious Mocha has been added
- Bewitched Black has been added
- Party pink is now much less pastel

File Changes:
- [Camera Manager].prefab: Now has a color blindness filter on it. Turn it on to see the game screen in there perspective of different types of color blindness.
- Materials: All emissive materials have been set to baked emission to help performance. This will probably need to be tweaked later but I noticed the game going from around 100fps to 150fps when emission was set to baked. I didn't notice a visual difference though.

How to test:
- Check out the Colorblind Test scene. Press play. On the camera, find the color blindness component and switch the filters. Each character color should be distinctive on each filter.
- Play the lobby. Keep changing your character color. You should notice new black and brown colors. Pink and orange should look different. You shouldn't see yellow and red anymore as options.

Notes:
- This doesn't solve our color blindness problem. We likely need different texturing and accessories to truly distinguish our characters. This is a step in the right direction though.


JamesCoulter:
Maybe a way to help the black and brown stand out might be to make the skeleton brighter, or perhaps using a shader to make a Tron-esque border glow? They could be thought of as special skins alongside that "vaporwave" one you made.

PR #197, opened 2019-05-16
Cross bomb to time bomb

JamesCoulter:
Contributors: Coulter

Goal:
Add a radial timer bar onto the cross bomb block to provide a better indication of how much time is left before it explodes.

Files Changed:
- BombExploder.cs: Added a function in Update() with a bit of math that continually reduces the radial bar based on the bomb block's countdown timer.
- CrossBombBlock.prefab: Created new bomb block with a canvas containing the radial timer image components.
- Spawner.cs: Replaced original cross bomb with new one where necessary.

How to test:
- Either wait for a bomb block to spawn, or just put one in the scene so you don't have to wait.
- The red ring around the bomb sprite now gets reduced in a clockwise direction. Observe this, and make sure that the ring is completely gone as soon as the initial countdown ends.
- Push the bomb block. The ring should disappear instantly.
- Set the timer variable to different values, and repeat the above steps for each. No matter what the value is, the bar should always start completely filled, and take the exact amount of time you set to become completely empty.

Notes:
Right now the radial timer goes until the point where the warning cross is activated. If you would prefer that it go all the way until the final explosion, across both the initial countdown and the 1-second build-up to explosion, it's an easy fix to implement. Also, I am currently researching a convenient way to make the ring glow to be a bit more noticeable, but I personally think the radial bar is mostly visible enough, so I didn't feel it was necessary to include in this pr. If you think it is necessary, or if you think I should just not worry about it, let me know.


Assets/Scripts/Blocks/BlockExploder.cs
```diff
@@ -2,20 +2,31 @@
...
+    private GameObject _radialTimer;
+
     [SerializeField]
     private float _timer = 10f;
 
     [SerializeField]
     private float _timeBeforeDetonation = 1f;
 
+    private Image _radialBar;
+    private float _radialAmountFilled = 100f;
```
IanGRap: can we set the _radialAmountFilled = 10 * _timer. That way it will scale if we change the time value

JamesCoulter: I implemented this, but I also had to put in a bit more math to make it work properly. Added a "normalizer" variable to make sure the fillAmount and and the rate at which it decreased scaled correctly.

AustinKellar:
Um, I may be blind here, but I'm not seeing anything. I'll try again tomorrow on my other computer.

JamesCoulter:
@AustinKellar So either your branch pull is messed up, or the bar is just not that visible to you. If it's the second case, that means it's not noticeable enough. Problem is I'm not quite sure how noticeable it _should_ be. Making the bar glow would be extremely difficult from what I've read, so the alternative is making the bar wider, probably?

AustinKellar:
@JamesCoulter I think it was likely my pull. I want to try again tomorrow on a different computer.

PR #208, opened 2019-05-19
Respawn block revision 2

IanGRap:
Contributors: Ian

Goal: Create a more elaborate UI that helps the player track when they get squished to when they respawn as the respawn block.

Important Files:
- RespawnImage.cs: This file tracks the respawn image. The new updates include the image appearing when the player died when they are squished, having a radial timer that indicates how long it is until the respawn block spawns and panning from the player's death position to the top of the screen where the respawn block is coming from.
  - PanSpeed is the speed at which the image moves from the player's death position to the top of the screen.
  - EdgeBuffer is the percentage of the screen that is buffered so that the respawn image doesn't go to the edge of the screen and get cut off.
  - Transparency is the transparency of the icon (not determined by the players color).
  - FadeInTime is the time it takes the image to fade from full transparency to the transparency assigned.
  - TweenSize determines if the image tweens its size on a loop when it moves to the top of the screen. This is to make the image a bit more noticeable.
  - The position image function takes in world coordinates and converts them to screen coordinates. The code is a bit confusing so the reference source is listed in the comments. If either sets the image's position to where the player dies, or to where the respawn block will be coming if it has spawned.

- Spawner.cs
  - Players to respawn is now a list of respawn queue info. This includes player data and the corresponding respawn image icon. When the player dies, a respawn image is created and assigned to the position they died. When the respawn block is spawned, the image is updated with the respawn block's transform.
  - TimeTillNextSpawn is used by the RespawnImage for the radial timer

- RespawnUI.cs
  - Most of the code was deleted or moved to respawn image.
  - It now basically just spawns respawn images and assigns them as children to its canvas

How to test: Play a few rounds on the match/sprint scene. When players get squished, the UI icon should pop up in that location. There should be a circle bar around the icon the decreases. When it is done the icon should have an arrow appear then move to the top of the screen where the respawn block is coming from. Do these changes feel like they help you track being squished to becoming a respawn block. Do you think these are enough additions to keep players paying attention after they have been squished?


Assets/RespawnImage.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    // Code reference: https://answers.unity.com/questions/1037969/arrows-pointing-to-offscreen-enemy.html
+    /// <summary>
+    /// Take a position in the world and convert it to a screen position for the icon
+    /// </summary>
+    private void PositionImage()
+    {
+        if (!_transformAssigned)
+        {
+            Vector3 screenPos = _camera.WorldToViewportPoint(_startPosition);
```
AustinKellar: Woah. Math.

Assets/Scripts/Blocks/BlockSpawner/Spawner.cs
```diff
@@ -390,4 +396,16 @@ private class DelayedBlockFormation
         public int requiredWallCompactCount;
         public Transform formation;
     }
+
+    private class RespawnQueueInfo
```
AustinKellar: Could we just add RespawnImage to PlayerData instead of making a new object? Its kinda nice having it all in one class.

JamesCoulter: Sort of inclined to agree, mainly just because Spawner.cs already has way too much stuff in it.

IanGRap: I can do that

JamesCoulter:
The image still looks too big in my opinion, but as long as it's easily scalable it's not a big deal.

IanGRap:
Can you post a picture of how it looks now

JamesCoulter:
This is how it looks currently for me:
Screenshot (126)

If I reduce the scale values to 1, it looks like this, which I personally like a lot more:
Screenshot (127)

JamesCoulter:
But apparently it's already larger on my computer screen, so I wouldn't worry about it too much

IanGRap:
okay It looks like my changes are working. I'll drop the scale to 1 then push. Also you are in free aspect, I recommend choosing 16:9

JamesCoulter:
16:9 didn't have much effect in this particular case, but still a good idea, so thanks for the reminder. If you're lowering the scale, the edge-screen buffer should be changed too, in case you weren't doing that already.

PR #209, opened 2019-05-19
New Options Art

AustinKellar:
Contributors: Austin

Goal:
- To import the new options block art and slightly modify the functionality

What Changed:
- Camera position changed for options menu
- New 3D model and UI layout for options menu
- UIOptionsMenu
    -Changed the functionality such that you have to press the apply button before any settings are changed and you scroll left and right through all of the options, rather than pressing A for some. 
-  GameSettings.cs / ExtensionMethods.cs
    - Added a new particles setting called Party :). It scales up particle emission rate by 3x. The default value is High but I thought this setting could be fun. If y'all don't like it, its easy to undo.

How to Test:
- Load up the lobby scene and make sure that the functionality changes work:
    - You need to press Apply before any settings take effect
    - You need to scroll left and right to select any of the options (instead of pressing A)


Assets/Scripts/Public Helper Classes/GameSettings.cs
```diff
@@ -122,4 +122,4 @@ private class SaveData
 }
 
 public enum AntiAliasingOption { Off=0, Two=2, Four=4, Eight=8 }
-public enum ParticlesOption { Off=0, Low=6, Medium=3, High=1 }
+public enum ParticlesOption { Off=-1, Low=6, Medium=3, High=1, Party=0 }
```
IanGRap: I'm curious, why these number choices on the options?

Assets/Scripts/UI/Lobby/UIOptionsMenu.cs
```diff
@@ -162,35 +171,42 @@ private void SetAntiAliasingText()
...
 
     private void ToggleFog()
     {
-        GameSettings.FogEnabled = !GameSettings.FogEnabled;
-        SetOnOffButtonText(_toggleFogButton, _fogPrompt, GameSettings.FogEnabled);
-        GameSettingsManager.Instance.ApplyFogSetting();
+        _selectedFogOption = !_selectedFogOption;
+        SetOnOffButtonText(_toggleFogButton, _selectedFogOption);
+        AudioManager.Instance.PlayOneShot("UI Menu Toggle");
+        ToggleApplyButton();
```
IanGRap: A lot of this code is reused, I'm wondering if there is a way to refactor it into one function where you pass in the option and button

Assets/Scripts/Public Helper Classes/GameSettings.cs
AustinKellar: Yeah, its a little weird but I couldn't think of a better system. The way the extension method works is it scales down the emission rate according the integer value on the enum. Off and Party are non-positive to indicate that they work differently than the other ones. Off is ignored and Party has a multiplicative effect. Also, enums are ordered based on their values, so I have them going in reverse order: 6 --> 3 --> 0 --> -1,    (Low --> Medium --> High --> Party --> Off). The whole thing would be a lot more convenient if I could use a Struct instead of an enum (but the json writer cannot translate a struct within a struct). It would also be a lot more convenient if enums could take floats as values, but they can't.

Assets/Scripts/UI/Lobby/UIOptionsMenu.cs
AustinKellar: Sounds good. I'll look into that.

IanGRap:
Also, it would be nice to have a b for back prompt on the menu

AustinKellar:
Hmm, I'm not sure. I'll try to get to that if we have time at the end of the sprint. It's not exactly trivial but I don't think it would be too hard.

PR #211, opened 2019-05-19
Add new sounds

AlfredoCuevas:
Contributors: Alfredo, 

Goal: Add the Score Laughing Sound. Add the Transition sound when you click Play and go to the Lobby

Files Changed: 
- VariableSoundPitchShifter.cs: Created a new function that makes it easier to add new sounds groups. Added a sound group for the score Increment laughs and the respawn block break
- AudioManager.cs: Has a new function that will call the ScoreIncrement laugh sounds and the respawn block sounds. The pitch for this group of sounds can be adjusted in the Inspector but I left them with no pitch shift. 
- UIScoreImage.cs: calls the laughing sound when the score is updated but only if the score canvas is active.
- LobbyStateManager.cs: calls play on the transition sound. 
- PlayerRespawn.cs: Added the sound for when a respawn block breaks

**Note**: The new sound sources are not hooked up to the mixer. This should be done later when Fernando is remixing the music. 

How to test: 
* Start in the Lobby Scene. When you click the play button you should hear the transition. Try to spam this button by going into and out of the Lobby. Make sure that this still sounds OK.
* Play through the game until you get the score screen. Make sure that the score increment sound is called at the right time. (right when the score goes up in value)
* While you play make sure that the laugh sound is not played at any other time. I had an issue where the sound was being played even after the score screen was gone because it was updating the numbers in the background.
* Destroy respawn blocks. Both kill a player with one and destroy it normally. 
❤️
--- 
I updated variableSoundPitchShifter.cs with suggestions provided by ian.


Assets/Scripts/Audio/VariableSoundPitchShifter.cs
```diff
@@ -9,6 +9,12 @@ public class VariableSoundPitchShifter : MonoBehaviour
     [SerializeField]
     private SoundGroup _blockBreak;
 
+    [SerializeField]
+    private SoundGroup _scoreIncrement;
+
+    [SerializeField]
+    private SoundGroup _respawnBlockBreak;
+
```
IanGRap: Making a specific variable for each of these feels like it will be overwhelming soon (or at least it would if we didn't have to be done in like 2 weeks). I feel this should be a list of sound groups and each sound group should have a string name.

Assets/Scripts/Audio/VariableSoundPitchShifter.cs
```diff
@@ -23,22 +29,39 @@ private void Awake()
...
+    public void PlayOneShotScoreIncrement(bool pitchShiftOn = true)
+    {
+        PlayOneShotSoundGroup(_scoreIncrement, pitchShiftOn);
+    }
+
+    public void PlayOneShotRespawnBlockBreak(bool pitchShiftOn = true)
+    {
+        PlayOneShotSoundGroup(_respawnBlockBreak, pitchShiftOn);
+    }
```
IanGRap: Tying into my above comment, these functions should be 1 function where they search the list of sound groups for a name then call PlayOneShot on the found sound group.

Assets/Scripts/Audio/VariableSoundPitchShifter.cs
```diff
@@ -23,22 +29,39 @@ private void Awake()
...
+        while (index == soundGroup.PreviousIndex)
         {
-            index = UnityEngine.Random.Range(0, _blockBreak.sounds.Length);
+            index = UnityEngine.Random.Range(0, soundGroup.sounds.Length);
         }
 
         if (pitchShiftOn)
         {
-            _blockBreak.sounds[index].pitch = UnityEngine.Random.Range(_blockBreak.pitchMin, _blockBreak.pitchMax);
+            soundGroup.sounds[index].pitch = UnityEngine.Random.Range(soundGroup.pitchMin, soundGroup.pitchMax);
```
IanGRap: should the pitch be reset after? What if I call a sound group with pitch shift on and then call it again with it off, won't the pitch shift be set at the random level from the previous call?

Assets/Scripts/Blocks/PlayerRespawner.cs
```diff
@@ -48,7 +48,12 @@ public override void Break(bool playSound = true)
             respawnBeam.transform.SetParent(player.transform);
             respawnBeam.GetComponentInChildren<Renderer>().material = blockMaterial;
         }
-        base.Break(playSound);
+
+        if (playSound)
+        {
+            AudioManager.Instance.PlayOneShotRespawnBlockBreak();
+        }
+        base.Break(false);
```
IanGRap: Why don't we just set the break sound of the respawn block to be the respawn block sound. This code seems unnecesarry.

Assets/Scripts/Audio/VariableSoundPitchShifter.cs
AlfredoCuevas: For some reason, I thought that because SoundGroup was a custom class that a list wouldn't show up in the Inspector. I agree thought that this should be a list but didn't test to make sure it would work. Ill try to change it later.

Assets/Scripts/Audio/VariableSoundPitchShifter.cs
AlfredoCuevas: This shouldn't matter because these sound groups are only meant to pitch shift sounds that have lots of variation in them. In the AudioManager I don't give you an option to call these functions with pitch shift off. You can only set the pitch shift in the inspector and that applies to any time the sound group is called.

Assets/Scripts/Blocks/PlayerRespawner.cs
AlfredoCuevas: I'm not sure what you mean here. This sound is calling the RespawnBlock sound. Because this file inherits from the BlockDamager script it needs to call base.Break(). However, this has a call to the regular block break sound in it. So I had to call false _base.Break(false)_ in order to turn off that block break sound. THen this if statement just calls the respawn block breaks sound.

Assets/Scripts/Audio/VariableSoundPitchShifter.cs
IanGRap: Yeah I've noticed with Unity that basically anything can be visible in the inspector as long as it isn't a key value pair.

Assets/Scripts/Audio/VariableSoundPitchShifter.cs
IanGRap: works for me

Assets/Scripts/Blocks/PlayerRespawner.cs
IanGRap: If you look at the block damager script in the inspector, you can set what the break sound is. You can change it from block break rock to respawn block. However, now I see that it is a group. If you called groups with string idea using the idea listed above, this would still work.

AlfredoCuevas: If you look in the code the string is no longer used to call the block break sound. Ill see about changing it after I refactor the code based on your suggestions above.

IanGRap:
Good stuff!

PR #212, opened 2019-05-20
Respawn block input revision

IanGRap:
Contributors: Ian

Goal: Refine the block input so it feels responsive.

Important Changes:
- BlockInput.cs: Removed the last frame check. This appeared to be what was causing the jankiness on the inputs. Push threshold requires the joy stick being pushed a certain amount in a direction before being registered. Up pushing requires an additional flag to be set to true. Down pushing has been removed (for now).
- BlockMovement.cs: Set values from private to protected/virtual so they can be overriden. The assign variables function is so that it can be overridden, avoiding overriding unity functions. Apply gravity is now lerp.
- Spawner.cs: Searches for a MeshRenderer instead of Render for setting color so it doesn't change colors of particle effects.
- RespawnBlockMovement.cs: Destroys the block when it lands if the flag is set. Overrides gravity so that if the player is pressing down, they speed up the fall rate of the block.
- AudioManger.prefab: Added block move sound.

How to test: I made a block test scene that should make it easier to mess around with the block input. Get squished and try moving around as the respawn block. It should feel responsive. You can also use the dpad. If you press down, you will speed up instead of pushing down. There should be a subtle fall line particle effect to let you know you are accelerating down.

Notes: You can't push down currently, but contradicts with being able to speed up moving down. This addition will probably need a new pr.


Assets/Scripts/Blocks/RespawnBlockMovement.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        if (_breakOnLanding && _blockAdjacencyQueries.FallingPathBlocked())
+        {
+            _blockDamager.Break();
+            return;
+        }
+
+        if (!_blockPusher.IsPushed && _blockInput.InputDown)
+        {
+            _rigidbody.velocity = Vector3.Lerp(_rigidbody.velocity, Vector3.down * _terminalVelocityWhileHoldingDown, _accelerationWhileHoldingDown * Time.fixedDeltaTime);
```
AlfredoCuevas: I'm curious why you are using Lerp here

Assets/Scripts/Blocks/RespawnBlockMovement.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System.Collections;
```
AustinKellar: Also just curious. We do we need a separate script for RespawnBlockMovement? Is it just for destroying it when its grounded?

Assets/Scripts/Blocks/RespawnBlockMovement.cs
IanGRap: If the rigidbody has sped up by pressing down but you let go it needs to slow back down

Assets/Scripts/Blocks/RespawnBlockMovement.cs
IanGRap: No it also allows you to speed up the acceleration if you are holding down. The speed you go is faster than normal acceleration.

PR #213, opened 2019-05-20
Issue 210

IanGRap:
Contributors: Ian

Goal: Make it so that the player doesn't get darker when they are cooling down from a dash. Instead they have a volumetric light that turns off. Also added scripts to ensure volumetric lights are turned on.

Important Files:
- TurnAuraOn.cs: Turns on the aura main component on the camera when awake. We've been having problems were these components would randomly be turned off on play.
- TurnAuraLightOn.cs: Turns on the aura light component on start. This is on start so we know that the AuraMain component has already been turned on (if the main component isn't on, then none of the other components will turn on).
- PlayerDash.cs: Turns the aura light component on and off when the dash is ready and not ready respectively.
- Player.prefab: Character color now doesn't change when the player dashes. Instead, their volumetric light turns on and off.
- Sprint 6.unity and Win Scene.unity: Added the aura light component to the background spotlights.
- [Camera Manager].prefab: Added aura on component.
- Death Block.prefab: Added aura light on component.

How to test: Play a match. The players should have a glow that will turn off when they dash. It will turn back on when they are ready to dash again. The background spotlights should always be on. The aura on the death block should always be on. How does the aura on the player look? Is it too much? Is it an aesthetically pleasing addition. It should make players more visible.


JamesCoulter:
I feel like you're okay to remove everything related to the material color change, since it seems like that idea was a bust. Unless we still want to include it as a really minor color change. But it seems like we don't, so it's not really needed anymore.

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -59,6 +59,7 @@ public class PlayerDash : MonoBehaviour
     private PlayerGrounded _playerGrounded;
     private Animator _animator;
     private Light _playerLight;
+    private AuraAPI.AuraLight _auraLight;
     private SkinnedMeshRenderer _playerRenderer;
     private Material _playerMat;
```
JamesCoulter: _playerRenderer and _playerMat aren't really being used anymore

Assets/Scripts/Player/PlayerDash.cs
```diff
@@ -164,6 +166,7 @@ private void Dash()
             if (_enableDashColorChange)
             {
                 _playerLight.enabled = false;
+                _auraLight.enabled = false;
                 _playerRenderer.material.SetColor("_Color", _dashCooldownColor);
```
JamesCoulter: Neither of these SetColor() lines are being used anymore

PR #214, opened 2019-05-22
Lobby Color Change Rework

AustinKellar:
Contributors: Alfredo, Austin

Goals:
- Get rid of paint blocks (and code)
- Create new system for changing color on Dab
- Reorganize lobby hierarchy
- Fix the font bug in Begin Rave screen.
- Changed "Begin Rave" to "Ready to Rave?" in the UI

IMPORTANT NOTE:
- Right now, we don't have a finished Y button icon. We will make a tiny PR to put it in once the asset is finished.

What Changed:
- Got rid of PaintBlockCollision, PlayerSelectBlockSpawner, PaintBlockMovement
- Added a new script called PlayerColorChanger, which changes player colors and spawns the visual effects
- Changed PlayerSelectColorController so that it ignores players' preferred materials on the color change dab
- Added a property called CanTaunt to PlayerTaunt.cs so that the color changer does not initiate the effect if the player is already dabbing
- Reorganized scene hierarchy
- Changed UI text and font in the Begin Match scene
- New prefab called Color Change Beam which is the spotlight visual effect
- New particle effect called Color Change Particle Effect

How to Test:
- Load up the Lobby scene
- Make sure the hierarchy is clean (before all of the title screen objects were spread all over the hierarchy rather than being consolidated to a parent object)
- Make sure that when your player is in the lobby, you can change colors with Y. 
    - You cannot change colors if you are already in the animation
    - You cannot change colors if you are on the ready stage
    - You cannot change colors when you are in the air (the conditions are the same as the dab conditions)
- Make sure that the icon colors are right in the Ready to Rave scene
- Make sure everything still works in a build

- Do you like the VFX?


Assets/Scripts/Player Selection/PlayerColorChanger.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private IEnumerator DoColorDab(GameObject player)
+    {
+        yield return new WaitForSeconds(_colorChangeDelay);
+
+        if (player != null)
+        {
+            Material currentMaterial = player.GetComponentInChildren<SkinnedMeshRenderer>().material;
+            _colorController.AssignRandomPlayerMaterial(player.gameObject, false);
```
IanGRap: I feel like instead of random this should be next color. While playing, it took me a real long time to get to blue. If I wanted blue I feel like this would be frustrating.

Assets/Scripts/Player Selection/PlayerSelectColorController.cs
```diff
@@ -37,16 +30,26 @@ public bool MaterialInUse(Material material)
...
-        Material material = GetPreferredOrRandomMaterial(_playerMaterials[index]);
+
+        Material material;
+        if (usePreferredMaterial)
+        {
+            material = GetPreferredOrRandomMaterial(_playerMaterials[index]);
+        }
+        else
+        {
+            material = GetRandomAvailableMaterial();
```
IanGRap: Again, can this material instead be the first one in the list and when you change materials your old one becomes the last one in the list?

Assets/Scripts/Player Selection/PlayerColorChanger.cs
JamesCoulter: Agreed

PR #220, opened 2019-05-24
Updating materials manager

IanGRap:
Contributors: Ian

Goal: Refactor how character materials were implemented to make room for new, more complicated skelegoo skins.

ImportantFiles:
- Skin.cs: Stores info for goo, bones and eye socket materials, along with dominant color, a UI image and a name. Also has static function for assigning a character a skin.
- Spawner.cs, PlayerColorController.cs, WinScreenController.cs, RespawnUI.cs, PlayerColorChangers.cs, PlayerSelectColorController.cs, PlayerSpawner.cs, PlayerData.cs (if I didn't list the change for a file, it probably falls under this category): Removed referencs to player material and updated them with player skins.
- PlayerRespawner.cs: Now has a reference of the player skin it will be applying it respawns. Setting the player skin also sets the materials on the respawn block.
- Materials.cs: Now tracks skins instead of just materials. Has functions to search for skins by color and by name.
- PlayerDash.cs: Removed option of changing color on dash. This was causing problems with the skin changes.

How to test: Run though every scene. You shouldn't get any errors. Your colors should carry from scene to scene. The UI Icons should all be appropriately colored. You should be able to change colors and back out on the lobby scene. The respawn blocks to should update their colors properly (The blockinput test scene is very good for testing this). On the materials manager, set the colors, icon and materials of a skin to random weird stuff. These changes should appear with characters. Test in build.

Notes: It looks like a lot of code, but its mostly just replacing material with skin.


Assets/Scripts/Blocks/PlayerRespawner.cs
```diff
@@ -24,28 +28,47 @@ private void Update()
...
+    {
+        if (_playerSkinAssigned)
+        {
+            return;
+        }
+
+        _playerSkin = skin;
+        _playerSkinAssigned = true;
+
+        var renderers = GetComponentsInChildren<MeshRenderer>();
```
AustinKellar: Our norm has been to not use var. is this type really long? Why did you go with var?

Assets/Scripts/Player Selection/PlayerSelectColorController.cs
```diff
@@ -7,8 +7,8 @@
 
 public class PlayerSelectColorController : MonoBehaviour
 {
-    private List<Material> _materialsInUse = new List<Material>();
-    Dictionary<PlayerIndex, Material> _playerMaterials = new Dictionary<PlayerIndex, Material>
+    private List<Skin> _skinsInUse = new List<Skin>();
```
AustinKellar: Man, this must've been fun lol. Logic looks almost the same but almost everything is renamed lol. Nice.

Assets/Scripts/Player Selection/Skin.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+[System.Serializable]
+public class Skin {
+    public string name;
+    public Material gooMaterial;
```
AustinKellar: Nice.

Assets/Scripts/Player Selection/Skin.cs
```diff
@@ -0,0 +1,36 @@
...
+    public Material skullMaterial;
+    public Material eyeSocketMaterial;
+    public Sprite image;
```
AustinKellar: I would prefer if these were public gets and private sets. Do these need to have public setters?

IanGRap:
yeah no new skins

Assets/Scripts/Player Selection/Skin.cs
```diff
@@ -0,0 +1,49 @@
...
+public class Skin {
+    [SerializeField]
+    public string _name;
+    [SerializeField]
+    public Material _gooMaterial;
+    [SerializeField]
+    public Material _skullMaterial;
+    [SerializeField]
+    public Material _eyeSocketMaterial;
+    [SerializeField]
```
AlfredoCuevas: Only nitpick I have is that the variables here feel a little packed. I think we have been putting a space between the names and [SerializeField]. Or you could also put the variable on the same line as its [SerializeField].

PR #221, opened 2019-05-24
Light on respawn and bomb

JamesCoulter:
Contributors: Coulter

Goal:
Replace the programmer art on the respawn and bomb block beams with volumetric lighting.

Files Changed:
- PlayerRespawnBeam.prefab: It seemed a bit odd to me that Destroy() was called from a child, and some refactoring was done to move functionality to a script on the parent. A volumetric light was added as a child and its functionality is controlled in the parent.
     - PlayerSpawnBeamController.cs: Since another object is childed to the parent beam object, it made sense to me to refactor a bit so that a script on the parent controls the behavior of the children.
     - PlayerSpawnBeam.cs: This script now only controls the trigger interactions for the block it's attached to.
- CrossBombBlock.prefab: Added volumetric lights for the warning effect.
     - ExplosionWarningController.cs: Activates the volumetric lights and tweens them.
- CrossExplosion.prefab: Added volumetric lights for the explosion effect.
     - DestroyOnTimer.cs: Activates volumetric lights and tweens them.

How to test:
- Die
- Kill someone as a respawn block. Observe the respawn beam. Do you like the appearance/intensity?
- Respawn different color players and make sure the beam is always the correct color.
- Observe the warning and explosion effects on the bomb block (putting one in the scene will help for efficient testing). Do you like the appearance/intensity of the effects?

Notes: 
There was some Aura weirdness that I couldn't resolve. For example, I can't get the explosion warning light to show through other objects without looking disjointed. The warning lights in general are also weirdly fuzzy. The only solution as far as I can tell is upgrading to a paid version of Aura with more functionality and specialization options.


Assets/Scripts/Other/DestroyOnTimer.cs
```diff
@@ -12,26 +12,21 @@ public class DestroyOnTimer : MonoBehaviour
...
-        }
-        foreach (var light in _lights)
+        foreach (Transform child in transform)
         {
-            light.DOIntensity(_intensity, _timer);
+            if (child.GetComponent<AuraAPI.AuraLight>() != null && child.GetComponent<Light>() != null)
+            {
+                child.GetComponent<Light>().DOIntensity(_intensity, _timer);
+                child.GetComponent<AuraAPI.AuraLight>().enabled = true;
+            }
```
IanGRap: good stuff

PR #222, opened 2019-05-25
Lobby Changes Round 2

AustinKellar:
Contributors: Austin

Goal:
- Place blocks inside lobby
- Implement an exit that allows players to despawn and return to the title menu

Note:
- Exit sign art and Change Color icon is still temp

What Changed:
- LobbyBlockSpawner:
    - A simple block spawner that spawns when there are fewer than 3 blocks (this value is configurable)
        - Can toggle off respawning with a checkbox if we so choose
- LobbyBlockDestroyer
    - A simple script that goes on a trigger collider which destroys blocks upon touch
        - Used to make sure blocks don't get on the ready stage and the exit area
- SquishDetector, BlockPusher, PlayerAdjacencyTracker, BlockDamager
    - Made small changes so that blocks can be pushed in the Lobby
- LobbyStateManager, PlayerSpawner
    - Added functionality so that the person who presses Play spawns into the lobby
- PlayerExitCollider
    - Despawns players and returns to title menu if the last player leaves the lobby

How to Test:
- Load up the lobby
- When you press play, you should spawn in
- Try pushing blocks around. They should behave as normal
- When there are 3 or fewer blocks in the scene, more should spawn in
- Try dying. You should respawn
- Try going to the exit. If there is more than 1 player, you should be despawned. If you are the last player, you should despawn and be taken back to the title screen
- Test in a build.


Assets/Scripts/Blocks/BlockPusher.cs
```diff
@@ -97,11 +97,24 @@ public virtual bool Push(Vector3 direction, float animationDelay = 0f)
...
+                from obj in adjacentObjects
+                where obj.tag == "Block"
+                select obj.GetComponent<BlockPusher>()
+            ).ToList();
+        }
+        else
+        {
+            adjacentBlocks = (
+                from obj in adjacentObjects
+                where Spawner.Instance.BlockTagList.Contains(obj.tag)
```
IanGRap: maybe these scripts should have a privately assigned list of block tags that defaults to just block if the spawner is null. I don't think we have time for the refactor though.

Assets/Scripts/Player Selection/LobbyBlockSpawner.cs
```diff
@@ -0,0 +1,55 @@
...
+    private GameObject _blockPrefab;
+
+    [SerializeField]
+    private bool _blockRespawningEnabled = true;
+
+    [SerializeField]
+    private int _respawnThreshold = 3;
+
+	private void Start ()
+    {
```
IanGRap: awkward curly brace

Assets/Scripts/Player Selection/LobbyBlockSpawner.cs
```diff
@@ -0,0 +1,55 @@
...
+        StartCoroutine(SpawnBlocks());
+	}
+
+    private IEnumerator SpawnBlocks()
```
IanGRap: should this coroutine even be called if _blockRespawningEnabled is false?

Assets/Scripts/Player Selection/LobbyBlockSpawner.cs
```diff
@@ -0,0 +1,55 @@
...
+                SpawnNewFormation();
+            }
+
+            yield return new WaitForSeconds(5f);
+        }
+    }
+
+    private void SpawnNewFormation()
+    {
+        GameObject tempFormationObject = Instantiate(GetRandomBlockFormation());
```
IanGRap: you dont need to instantiate the outline object. You can run the forloop on the prefab

Assets/Scripts/Player Selection/LobbyBlockSpawner.cs
AustinKellar: It looks right in VS. I think its GitHub's interpreter being awkward.

Assets/Scripts/Player Selection/LobbyBlockSpawner.cs
AustinKellar: Good point.

Assets/Scripts/Player Selection/LobbyBlockSpawner.cs
AustinKellar: Fixed.

Assets/Scripts/Blocks/BlockPusher.cs
AustinKellar: Yeah, if we have time to refactor this, I will. I think its out of scope for this PR.

AustinKellar:
Alright, I implemented most of your suggestions. Block spawn locations are more contstrained and there are only 2 (small) block formations allowed. Since there is another lobby PR coming up, I will save all visual upgrades for that time.

PR #235, opened 2019-05-30
Improve death effect

JamesCoulter:
Contributors: Coulter, Liam, Reed, everyone else in the room

Goal: Redo the death particle effect so that it looks gooier/juicier and is less computationally expensive.

Files Changed:
- Player Death Particle Revisited prefab: New particle effect drawing some elements from the original, with new 2D droplets and splats used instead of 3D meshes.
- SquishDetector.cs: Rewrote lines that control particle color in order to work with the new 2D materials and the specific shader type they are using.
- ObjectSpawner.cs: Same as in SquishDetector.cs, so that the new effect is also used in the lobby.
- CameraGoo.cs: script now contains a serialized list of different screen splatter patterns, and the specific image used for each death is picked at random.

How to test:
- Kill players in both the lobby and match scenes. There should be a new goo explosion effect that is is the same color as the skin on the respective player. Different screen splatters, selected randomly each time, should appear on death.

Notes:
- In the code blocks that set the particle color, I included a link to the webpage that I used to figure out how to use the Material Property Block (which is necessary because of the specific particle shader type I am using). If you think the code is straightforward enough, I could remove the comment with the url link.


Assets/Scripts/Player Selection/PlayerSpawner.cs
```diff
@@ -115,17 +129,28 @@ public void Despawn(PlayerIndex index, bool removeActivePlayer)
             Transform playerTransform = player.transform;
             Destroy(player.gameObject);
 
-            ParticleSystem particleEmitter = Instantiate(_despawnParticleEffect, playerTransform.position, Quaternion.identity);
+            GameObject particleEmitter = Instantiate(_despawnParticleEffect, playerTransform.position, Quaternion.identity);
             particleEmitter.transform.parent = gameObject.transform;
 
             Skin skin = Materials.Instance.GetSkinByName(player.name);
-            particleEmitter.GetComponent<Renderer>().material = skin.gooMaterial;
+            foreach (ParticleSystem deathParticle in particleEmitter.GetComponentsInChildren<ParticleSystem>())
```
AlfredoCuevas: Maybe this foreach should become a function since we use it twice. Probably not too important since we only use it twice though.

Assets/Scripts/Player Selection/PlayerSpawner.cs
```diff
@@ -115,17 +129,28 @@ public void Despawn(PlayerIndex index, bool removeActivePlayer)
...
+                {
+                    //The block of code below was written based on the information at this link: https://thomasmountainborn.com/2016/05/25/materialpropertyblocks/
+                    _block = new MaterialPropertyBlock();
+                    _particleRenderer = deathParticle.GetComponent<Renderer>();
+                    _particleRenderer.GetPropertyBlock(_block);
+                    _block.SetColor("_TintColor", skin.gooMaterial.color);
+                    _particleRenderer.SetPropertyBlock(_block);
+                }
+            }
+            CameraGoo.Instance.ChooseASplat();
```
AlfredoCuevas: Nitpicking here. I would prefer if there was a space between this line and the ones below just to separate the code into chunks.

Assets/Scripts/Player/SquishDetector.cs
```diff
@@ -59,9 +61,22 @@ public void PlayerDeath()
...
+                {
+                    //The block of code below was written based on the information at this link: https://thomasmountainborn.com/2016/05/25/materialpropertyblocks/
+                    _block = new MaterialPropertyBlock();
+                    _particleRenderer = deathParticle.GetComponent<Renderer>();
+                    _particleRenderer.GetPropertyBlock(_block);
+                    _block.SetColor("_TintColor", material.color);
+                    _particleRenderer.SetPropertyBlock(_block);
+                }
+            }
+            CameraGoo.Instance.ChooseASplat();
```
AlfredoCuevas: Same here I would prefer a space after this line.

Assets/Scripts/Player/SquishDetector.cs
```diff
@@ -59,9 +61,22 @@ public void PlayerDeath()
...
+            foreach(ParticleSystem deathParticle in particleSystem.GetComponentsInChildren<ParticleSystem>())
+            {
+                deathParticle.PlayAccordingToGameSettings();
+                if (deathParticle.gameObject.name.Contains("Goo"))
+                {
+                    //The block of code below was written based on the information at this link: https://thomasmountainborn.com/2016/05/25/materialpropertyblocks/
+                    _block = new MaterialPropertyBlock();
+                    _particleRenderer = deathParticle.GetComponent<Renderer>();
+                    _particleRenderer.GetPropertyBlock(_block);
+                    _block.SetColor("_TintColor", material.color);
```
AlfredoCuevas: In the article it says that if you change a variable in a shader with this method and you don't add [PerRendererData] then Unity will continue to create new instances of materials whenever you change the value. Because you didn't create your own shader, I think Unity is still creating a new instance every time. I'm not sure how you would go about making a particle shader though, they seem to be different from normal 3d shaders.

JamesCoulter: I googled around for more info on [PerRendererData], and from what I can tell there's a lot of inaccurate info, including what's in the article I linked. A member of the Unity team replied on a forum post related to this guide that "The PerRendererData attribute is purely UI related and has no impact on performance." The article writer said he'd update it but it doesn't look like he ever did.

https://forum.unity.com/threads/learn-how-to-use-materialpropertyblocks-and-perrendererdata-for-great-performance-gains.406638/

Ultimately, It doesn't seem like it's something to worry about too much

PR #236, opened 2019-05-31
Credits

JamesCoulter:
Contributors: Coulter, Emma, Austin

Goal: Add an option to the start screen to see the credits.

Files Changed:
- Credit Block prefab: Holds the block asset, as well as all text objects displayed on it.
- LobbyStateManager.cs: Added in code needed to make the credits screen accessible.
- ControllerSignIn.cs: press B on credits to return to title screen.

How to test:
- Open Lobby scene
- Go to credits option and open it. It should be similar to opening the options menu.
- Press B to go back to the title screen.


AustinKellar:
Looks great.

PR #238, opened 2019-05-31
added slow motion effect to the last player death

AustinKellar:
Contributors: Austin

Goal:
- Add a slow motion effect to the last player death

What Changed:
- MatchManager
    - When there is 1 player left, time slows, then animates back to full speed over a couple seconds

How to test:
- Load up the match scene and play some games. 
- Make sure slow motion is triggered when someone wins
- Make sure time returns back to normal eventually
- Do you like the effect?


Assets/Scripts/Environment Actions/Managers/MatchManager.cs
```diff
@@ -127,4 +129,18 @@ public List<Vector3> GetPlayerPositions()
 
         return _alivePlayers.Select(p => p.transform.position).ToList();
     }
+
+    private void SlowTime()
+    {
+        float slowMotionSpeed = 0.25f;
+        Time.timeScale = slowMotionSpeed;
+        Time.fixedDeltaTime = slowMotionSpeed * 0.02f;
+        Invoke("ResumeNormalTime", 0.4f);
```
IanGRap: The time until normal time resumes should probably be a serialized variable

AustinKellar: Yeah, all of the numbers probably should be. Not sure why I didn't. Thanks.

PR #282, opened 2019-09-02
Old skins

IanGRap:
Contributors: Ian

Goal: Add in some of the old skins.

How to test: Go to the lobby and change your color. You should find colors for blue, pink, orange, gold, purple and white (The white one has an inverted skull color because I thought it looked cool) Each of these colors should have and image when you ready. Also load up the color blind test scene. All the colors should be distinctive under each of the color blind settings. You can change the color blind settings as the final component on the camera.


AlfredoCuevas:
I noticed that this branch has a conflict with the Editor.meta file. Feel free to keep the development version or your version of the file. Either way it should fix itself when Unity runs.

PR #288, opened 2019-09-23
Bd teams

IanGRap:
Contributors: Ian

Goal: Lay down initial code to set up Squish running as a team game.

Important Files:
-TeamStartingFormation.cs: A child class of starting formation. Has a list of player spawning positions for the opposing team.
- Skin.cs: Added new function that changes a player's skin color based on what team they are on.
- SceneCreator.cs: Added functionality for spawning a team formation if the game mode is block dunk.
- Teams.cs: A class that tracks team data for an A and B team. This includes a list of the player indices for each team, a team color and a team name. There is also a function that creates default teams.

How to test: Go into the Teams.cs file and change the arguments on line 42 and 43. If you change the default team colors, the player colors should change when you start the game. You can also try creating a TeamStartingFormation scriptable object and putting it in the scene creator.

Notes: Right now, it is assumed that a team will never consist of more than 3 players (you can do 3 vs 1 if you want) or less than 1 player when spawning players. It is also assumed we are using colors to distinguish teams, but this could change later. For example, hats could distinguish teams (an idea I kinda like). These would be changes for later pull requests.


IanGRap:
exactly

PR #290, opened 2019-09-28
Bd walls

IanGRap:
Contributors: Ian

Goal: Create the first pass of walls in block dunk. These walls are gaps with death blocks that spawn and break the floor to shrink the arena.

Important Files:
- ArenaEdgeBase.cs: The base class for inheritance for all compacting walls in the game. Tracks wall positions and an instance. The reason for the base class is that other classes can call the base for information, regardless if it is a compacting wall or the death blocks.
- DeathBlockEdgeManager.cs: Inherits from ArenaEdgeBase. Spawns death blocks in the game to reduce arena size.
- CompactorTimer.cs: A new class that monitors the timer for when the arena should compact. Also has the ability to decay the timer each compacting and set how many times the arena should compact.
- WallManager.cs: Removed timer functionality and put it into CompactorTimer.cs. Also set up functionality to inherit from ArenaEdgeBase.cs.
- BlockSpawnTimer.cs, BlockSpawner.cs, BlockClearer.cs, BlockSpawnZoneController.cs: Switch from referencing WallManager to referencing ArenaEdgeBase.
- DeathBlockCollision.cs: Added code so that only death blocks can break floor blocks.

How to test: Go to the block dunk scene and wait. Death blocks should spawn on the edges that will break the floor blocks. This should keep happening until the floor is 5 wide. Also play the normal match scene and make sure that everything works normally (especially the compacting walls).

Notes: I had a bunch of files that were changed that I didn't touch. Double check that there aren't any files changed that shouldn't be. Also mess around in the lobby and make sure everything seems normal. Also had weird changes there I had to roll back.


Assets/Scripts/Environment/Interactive/WallManager.cs
```diff
@@ -33,60 +23,34 @@ public class WallManager : MonoBehaviour
...
-            return;
         }
+        Instance = this;
 
         ErrorIfNull();
-        ResetCompactorTimer();
         Physics.IgnoreLayerCollision(_wallLayerNumber, _floorLayerNumber);
         Physics.IgnoreLayerCollision(_wallLayerNumber, _roofLayerNumber);
+
+        CompactorTimer.OnCompact += OnCompact;
```
AlfredoCuevas: I feel like these listener assignments should happen in Start(). Probably not a problem, especially because these are static events.

Assets/Scripts/Environment/Interactive/CompactorTimer.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System;
+
+public class CompactorTimer : MonoBehaviour
```
AlfredoCuevas: How modular is this piece from ArenaEdgeBase.cs? I think I would prefer that this script [Requres(...)] some form of the ArenaEdgeBase script, or maybe the reverse. The ArenaEdgeBase should Require this script. This would sort of couple them together. 

The script runs the risk of multiple versions of itself being present in the Hierarchy so coupling it would ensure that it can't exist in isolation.

Assets/Scripts/Environment/Interactive/BlockSpawnZoneController.cs
```diff
@@ -114,8 +114,9 @@ private void OnBreak_DecrementBlockCount(int count)
 
     public BlockFormation ChooseAShape()
     {
-        if (_hazardCooldown <= 0f && Random.Range(0f, 1f) <= (float)_blockCount / _highBlockCount)
+        if (_hazardBlockFormations.Count != 0 && _hazardCooldown <= 0f && Random.Range(0f, 1f) <= (float)_blockCount / _highBlockCount)
         {
+            Debug.Log("Choosing a hazard");
```
AlfredoCuevas: I don't think we need this debug statement.

Assets/DeathBlockEdgeManager.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class DeathBlockEdgeManager : ArenaEdgeBase
```
AlfredoCuevas: This script needs to be moved into the Scripts/ folder.

IanGRap:
All good calls

PR #292, opened 2019-09-29
Treasure grab part 1

JamesCoulter:
Contributors: Coulter

Goals: Implement treasure block with basic particles. Make sure that new point system works as desired.

Important Files:
- Treasure Grab Test.unity
- [Treasure Grab Match Manager].prefab: same as normal match manager object, with different scripts specifically for this mode.
- [Treasure Grab Block Spawner].prefab: same as the new match manager.
- Treasure Block.prefab: inherits from normal block scripts. gold in color. Has a boring particle effect.
- BlockCollisionDelegator.cs: PushBlock() function altered to also pass along a variable for the pushing player, if there is one)
- BlockPusher.cs: continues passing the above variable for the pushing player.
- TreasureBlockBreaker.cs: inherits from BlockDamager.cs, uses event to pass along info on pushing player and the amount of points to be given.
- GameManager.cs: Changed UpdateScore() to take in an argument for number of points.
- ActivePlayers.cs: Changed UpdateScore() to increase score by given value, rather than be hardcoded to increment by 1.
- TreasureGrabModeManager.cs: Checks for various events to issue score update commands, including giving players more points for breaking blocks and lowering their score when they die.
- GameState.cs: Add enum for treasure grab mode.
- PlayerRespawner.cs: Add bool so that if in treasure grab mode, players automatically respawn on hitting ground.

How to Test:
- Start game in Treasure Grab Test scene.
- Find and break treasure blocks. Watch console messages to make sure player scores are increasing and decreasing as they're supposed to. This means: Break a treasure block. Let a player with no points die. Let a player with points die. When players get squished, the squishing blocks directly adjacent to the player should turn into treasure blocks.

Notes:
I made this branch off of a previous commit after already completing the mode, because I did not plan out my implementation of the mode very well. I'm essentially breaking it apart to make the PRs more incremental and manageable, with the cost being that some things may not work properly quite yet (since the fixes might come in a later commit). Some issues, such as lack of UI and what happens when the game ends, are fixed later. I also encountered certain bugs in this version that I haven't noticed in the complete version of the mode, such as players respawning inside the block spawn zone. Another one that I only noticed once is blocks hovering in mid air after crushing a player. I'm not really sure why. Disregard for now, I think, and if they continue to be an issue in the later PRs for the mode, I'll resolve them then.


Assets/Scripts/Camera/CameraTransformTracker.cs
```diff
@@ -102,7 +102,7 @@ public void RemoveTrackedTransform(Transform removedTransform)
         if (_soloOnPlayerWin && !OnePlayerTracked)
         {
             int currentPlayerCount = _trackedTransforms.FindAll(t => t.tag.Equals("Player")).Count;
-            if (currentPlayerCount == 1)
+            if (currentPlayerCount == 1 && GameState.CurrentScene != CurrentScene.TreasureGrab)
```
IanGRap: you can just set _soloOnPlayerWin to false to avoid the camera focusing when there is one player left (if that is what you are trying to do here)

Assets/Scripts/Player/Movement/PlayerDash.cs
```diff
@@ -10,6 +10,9 @@
 [RequireComponent(typeof(PlayerGrounded))]
 public class PlayerDash : MonoBehaviour
 {
+    [SerializeField]
+    private GameObject _coinEffect;
+
```
IanGRap: it seems weird that this effect is in the dash scripts, but I can't exactly think of anywhere else to put it...

Assets/Scripts/Singletons/GameManager.cs
```diff
@@ -39,17 +27,14 @@ private void Awake()
         _matchEndManager = GetComponent<MatchEndManager>();
     }
 
-    public void UpdateScore()
+    public void UpdateScore(int points, PlayerIndex playerReceivingOrLosingPoints)
     {
-        ActivePlayers.IncrementScore((int)_winningPlayer);
-        if (ActivePlayers.GetScore((int)_winningPlayer) == _winsRequired)
+        ActivePlayers.UpdateScore((int)playerReceivingOrLosingPoints, points);
+        Debug.Log(playerReceivingOrLosingPoints.ToString() + ": " + ActivePlayers.GetScore((int)playerReceivingOrLosingPoints));
```
IanGRap: I noticed this debug coming up during survival match play

Assets/Scripts/Player/Movement/PlayerDash.cs
JamesCoulter: Yeah, that's the issue I was having. Nowhere really made sense to me

PR #293, opened 2019-09-30
Refactor lobby ui events

AlfredoCuevas:
Contributors: Alfredo

Goal: Make the UI elements in the lobby use the Unity UI Event system for easier modification in the future. Each "Screen" or UI section should be independent from the others. So Title Screen functionality is not tied together with the Settings through script. 

Changes: 
* **LobbyUIScreenManager.cs**: (Not a Singleton) This is what allows most of the UI to be navigated with Unity's Event System. (It is important that only 1 interactable UI screen is active/enabled at any given time.) This script handles the work of turning on an off a UI Game Object. Ex) From TitleScreen to Settings this script may be called by a button's OnClick() event to turn off the TitleScreen, Turn on the Settings screen, and select the first option. This script does _not_ handle camera movement. 
* **LobbyStateManager.cs**: Got rid of references to the Credits and Options menu. This script no longer handles turning them On or Off. Added some summaries to functions that I was constantly using. This script's functions are being used in the inspector when certain UI is clicked. 
* **UISettingsMenu.cs:** The replacement to UIOptionsMenu.cs. This now houses all of the functionality for controlling the Settings/Options menu. It is found on the Canvas of the Options Menu. It works in tandem with the Scrollbars to display and save the correct Settings. Note: The ApplyButtonClicked() function is used by the Apply button through the Inspector. 
* **ScrollBars**: So I had to do some hacky stuff with Unity Scrollbars to get the Options/Settings to look similar to how they did before. Because of how specific our settings are I couldn't think of a good way to generalize these into one script. So there are some made specific for certain options. 
* * UISettingsAntiAliasingScrollbar.cs: Used by the AntiAliasing scrollbar to make sure that the proper text is displayed and the correct setting is returned
* * UISettingsBoolScrollbar.cs: Used by any setting that can only be true or false. Displays the correct text based on the scrollbar value. 
* * UISettingsParticlesScrollbar.cs: Used by Particles scrollbar to make sure the proper text is displayed and the correct setting is returned. 
* * UISettingsResolutionScrollbar.cs: Used by the Resolution scrollbar " "
---
* **PlayerExitCollider.cs**: Added a UnityEvent that is called when the last player is despawned. It enables the use of LobbyUIScreenManager.cs from the inspector without having to hard code it into the script.
* **ControllerSignIn.cs**: Removed the input handling for OptionsMenu and Credits that was happening here
* **EventSystemInitializer.cs:** Used in the EventSystem Game Object. Finds the Input manager if for some reason the reference is missing after the scene is reloaded. 
* **UIEventTriggerFunctionWrapper.cs:** There was an issue where Objects that were DoNotDestroyOnLoad() would become missing references in the inspector. This script simply calls the AudioManager through script so that the Event Trigger component can still be used for buttons and stuff. 
* **UIOptionsMenu.cs:** I was kind of scared to delete any of the old scripts that controlled the UI. So I marked this one specifically as obsolete so that I wouldn't have to get rid of it or change it. This script however, should no longer be in use by the UI in the Lobby. 
* **GameSettingsManager.cs**: changed some string comparisons to use CompareTo() instead. Added a summary for myself. 
* **The Title Screen Menu Buttons**: use animations so that the Text can be customized to look like it did previously
* **[RewiredInputManager]**: Added new controller maps to register UI input
* **Lobby Duplicate.unity**: Just a copy of the old lobby in another scene, just in case

How to Test: 
* Test to make sure that you can still go through the whole game without it breaking. Example) Play > Select players > Finish a match > Go back to Lobby > Repeat again to test it works 2 times in one play session
* Go into the Settings. Does that work fine? Change some settings. Does it wall work correctly? Look out for things that aren't intuitive that I may have missed. 
* Do the same as above with the Credits.
* Make sure you can Quit
* The keyboard should still work as before for navigating the UI. Arrow keys for navigation. Right Shift for enter / A button. ESC for cancel / B button.
* Finally look through the hierarchy to see how I implemented the UI using the Inspector / Unity system. Does it make sense? Adding your own UI stuff should boil down to making a new Game Object with a canvas and knowing how to properly turn it on and off. 

Note: 
* I expect to have to change this before it's merged so don't be afraid to give feedback. 
* For me this was about learning how to use Unity system because I liked it more on paper. I am not personally attached to it. If people liked the old UI system better then I am ok with reverting back.
* I'm tired. 
* The ready screen (where it tells you to press Start to begin a match) was not modified. It had some custom stuff going and was located inside of one of the Singleton Managers. I figured it was alright to leave it alone since it was so independent from the other UI.


IanGRap:
I'm scared

AlfredoCuevas:
> I'm scared

Lol

AlfredoCuevas:
> Since this one is a biggie I haven't looked at the code yet. Here are my concerns from playtesting: I couldn't use my d-pad to navigate menus. When I am on a on/off option I can't keep pressing right to toggle throuhg options (Ihave to press right to go to on and left to go to off). The resolution option 3620x2036 doesn't fit on the button. Post processing effects are defaulted to off. When I return to the lobby, my cursor starts on options. Also, there is a merge conflict.
> 
> I do really like the little text animation on the main menu.

* The merge conflict makes me sad. Austin added the hats and changed the Lobby.unity scene. I don't think there is any way to merge that file. So I'm going to have to replace my Lobby with his and then completely redo all my changes. This sucks cuz this pull request was like 70% scene changes. 
* The D-pad shouldn't be a big problem. I should be able to add it to the controller map through Rewired, no code required. 
* The On/Off toggle is a limitation of using scrollbars. I wanted to copy the way Austin made the options. The scrollbar hack allowed me to keep the use of analog sticks for changing options. When you go "right" (On), what is actually happening is that the scrollbar handle is going all the way to the right. Making it wrap around is going to be tricky and probably not worth the time. 
  * One option would be to turn off the arrow image when you have reached the last option. So visually you know you can't toggle . Ex )   < On > becomes   < On           or         Off >
  * Alternatively, I could make the arrows buttons; under the hood there is an array of strings and the buttons just go through the array. This would mean you we cant simply move the analog right/left. The user needs to hover the right/left arrow button and select it with the A button. 
  * Option 3, break the consistency of the options menu a little. The on/off options can be actual toggles. A little box with a check mark symbolizing on or off.
* Ill look into making the space bigger for the resolution option that doesn't fit. 
* Post-Processing was already off by default. I didn't change the saved Settings.

AlfredoCuevas:
> Since this one is a biggie I haven't looked at the code yet. Here are my concerns from playtesting: I couldn't use my d-pad to navigate menus. When I am on a on/off option I can't keep pressing right to toggle throuhg options (Ihave to press right to go to on and left to go to off). The resolution option 3620x2036 doesn't fit on the button. Post processing effects are defaulted to off. When I return to the lobby, my cursor starts on options. Also, there is a merge conflict.
> 
> I do really like the little text animation on the main menu.

* I added the D-pad to the controller mapping so that should now work when navigating the UI
* I fixed the text for the Resolution so now it should't wrap and it should fit in the box
* The solution I went with for the options not wrapping when you cycle through them is that the arrows on the sides will dim when you reach the ends.

JamesCoulter:
I'm not sure if it's connected specifically to a change on this branch at all, but switching skins for players in the lobby doesn't work.

PR #295, opened 2019-10-01
Block Dunk: Scoring

IanGRap:
Contributors: Ian

Goal: Add scoring components that allow teams to win in Block Dunk. Also added a lot of new details to the goals.

Important Files:
- BlockDunkModeManager.cs: Tracks teams score and when to end the match based on a team's score.
- BlockDunkPointsIcon.cs: A UI world element that indicates how many points a team has.
- BlockDunkGoal.cs: Now displays an icon of the point value when a point is scored in the goal. The goals now move inward when a death block hits them and shrinks the arena. I also had to add a listener to wait for default teams being assigned to avoid one of those start vs awake race conditions.
- DeathBlockEdgeManager.cs: The edges of the arena are now tracked by the location of the goals.
- Teams.cs: Added ability to track wins of teams and resets score. I also added an event for after default teams are assigned to avoid race conditions when checking when players are assigned.
- UIScoreMenu.cs: Changed how score icons were removed and added to reflect teams, not players.
- PlayerCosmeticSetup: Assigns team colors for wins tracker UI.
- SceneCreator.cs: Had to move where skins were assigned to avoid race conditions where players weren't made yet.
- GameManager.cs: Added code to track if it is a team mode and check team wins for deciding a overall game winner.
- UIScoreImage.cs: Added functionality to assign a team instead of a player.

How to test: Load up the block dunk scene. Try pushing blocks into a goal. You should notice the UI at the top of the screen reflecting your score. You should also see a number pop up over the goal when you score points. The goals also have the basketball art! When a team hits the score limit, the round UI should pop up and note that a team has gotten more points. When one team hits the round win limit, you should move to the win screen.

Notes: Respawning and a team specific win screen are not updated yet. Those will be in later prs.


Assets/BlockDunkModeManager.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    private MatchEndManager _matchEndManager;
+
+    public int TeamAScore { get; private set; }
+
+    public int TeamBScore { get; private set; }
+
+    public int ScoreLimit { get { return _scoreCap; } }
+
+    public static BlockDunkModeManager Instance { get; private set; }
```
AlfredoCuevas: The Game Object that this script is attached too [Block Dunk Manager] has multiple scripts that are Singletons (they have a static Instance variable). I worry that there is a slim chance that they conflict with each other when on the same Game Object. 
Ex) What if I created another Game Object with BlockDunkModeManager.cs in it. Then it is possible that [Block Dunk manager] gets destroyed because when it gets to BlockDunkModeManager's Instance variable it is already set. That means I will have deleted the GameManager.cs script as well.

Assets/Scripts/Block Dunk/BlockDunkGoal.cs
```diff
@@ -1,13 +1,47 @@
...
+        if (Teams.TeamAIndicies.Count + Teams.TeamBIndicies.Count > 0)
+        {
+            SetColors();
+        }
+        else
+        {
+            Teams.DefaultTeamsAssigned += SetColors;
+        }
+
+        CameraTransformTracker.Instance?.AddTrackedTransform(transform.parent);
```
AlfredoCuevas: Why does this Instance have a ?

PR #296, opened 2019-10-09
Treasure Grab Part 2

JamesCoulter:
Contributor: Coulter

Goal: The PR has two main purposes. The first is to connect the treasure grab scene to the win screen, ensuring that the scene transition happens properly. The second main goal is to finish the treasure grab manager script and ensure all mechanics work properly. Later on, the PR was expanded to also restructure the point system to make it more easily applicable to all game modes.

Files Changed:
- ActivePlayers.cs: This one is fairly important. Players now have two separate scores - points and wins. Points are for any game mode where points are earned during matches, and wins are the number of matches won by a player. For example, In Treasure Grab, breaking treasure blocks earns points, and the first player to reach 10 points gets a win for that match.
- GameManager.cs: moved code that controls win scene transition out of UIMatchManager.cs and into GameManager.cs. This script also stores the variable for the number of match wins needed to win overall.
- TreasureGrabModeManager.cs: Contains the functions for giving players points from treasure blocks, and taking away points for death. When the needed number of points is reached, calls the MatchEndManager.cs functions. Debug line also added to track when players gain or lose points, since the UI for that will be in the next PR.
- [Treasure Grab Match Manager].prefab: This prefab is where the variables for number of points and number of wins can be changed. Ideally (for now), the Treasure Grab mode would just be a single match, so whomever gets one win first, would win overall.
- UIMatchManager.cs: Moved win scene transition code out of this script and into GameManager.cs. A future additional fix to consider would be doing the same for ends of matches.

How to test:
- Start up the Treasure Grab Test Scene.
- Look at the values for required wins and points in the treasure grab match manager.
- Based on that, start the scene and play. Once the correct number of points is acquired, the end of the match should trigger. Once that happens, the player with the winning number of points should receive the win. The needed number of wins will trigger the win scene transition. This should essentially all work the same as in the normal survival mode, the main difference being the addition of points that are acquired and lost during play.
- Observe the win scene. Make sure that the players are placed on the podium in the correct order.
- Go back and change the required point and required win variables to ensure that matches and games end as they should each time.
- Go to the Match scene and make sure the survival mode still works as it always has. I'm pretty certain my changes haven't messed up survival mode's functionality.

Notes: This should be the last larger PR for Treasure Grab. Future ones should all be significantly smaller. The next one is UI to track points during play, and after that will be things like bringing in the treasure chest model that Wesley made, and creating a visual of coins transferring to the player when a chest breaks. Future PRs will also be for tweaking gameplay after the finished mode gets some more feedback.


Assets/Scripts/UI/Match/RespawnUI.cs
```diff
@@ -20,6 +20,14 @@ private void Awake()
...
     }
 
+    private void Update()
+    {
+        if (GameState.MatchState == MatchState.MatchOver || GameState.MatchState == MatchState.GameOver)
+        {
+            gameObject.SetActive(false);
+        }
+    }
+
```
AlfredoCuevas: This isn't a huge problem but I also don't think it's insignificant. This feels like it should be some kind of event. Constantly checking just to deactivate this gameObject once feels inefficient. Can this be used with the OnMatchEnd() event inside MatchEndManager.cs?

Assets/Scripts/UI/Match/UITreasureScore.cs
```diff
@@ -0,0 +1,71 @@
...
+        foreach (UIScoreImage image in scores)
+        {
+            image.SetScoreText(ActivePlayers.GetWins(image.PlayerNumber));
+        }
+        if (GameState.MatchState == MatchState.MatchOver ||
+            GameState.MatchState == MatchState.GameOver)
+        {
+            gameObject.SetActive(false);
+        }
+    }
```
AlfredoCuevas: Where is this script being used? I couldn't find a reference to it in the Scene. If it is being used, I worry about this function being called in Update. If it's not being used I think it should be deleted.

Assets/Scripts/Blocks/Respawn/PlayerSpawnBeam.cs
```diff
@@ -10,14 +10,13 @@ private void OnTriggerEnter(Collider other)
     {
         if (other.tag == "Block")
         {
-            TreasureBlockBreaker treasureBlockBreaker = other.gameObject.GetComponent<TreasureBlockBreaker>();
-            if (treasureBlockBreaker != null)
+            if (other.gameObject.GetComponent<TreasureBlockBreaker>() != null || other.gameObject.GetComponent<PlayerRespawner>() != null)
```
AlfredoCuevas: The second bool check should be moved onto it's own line. It is a bit long. (Just to be clear: I think the code is fine. I only want some more white space)

Assets/Scripts/Blocks/Core/BlockDamager.cs
```diff
@@ -120,7 +120,7 @@ protected virtual void UpdateBlock()
         {
             DecrementHealth();
         }
-        if (HasKilledPlayer && GameState.CurrentScene == CurrentScene.TreasureGrab)
+        if (gameObject.GetComponent<TreasureBlockBreaker>() == null && HasKilledPlayer && GameState.CurrentScene == CurrentScene.TreasureGrab)
```
AlfredoCuevas: This bit of code worries me. `GetComponent<>()` is actually a decently expensive function call. On top of that `== null` null checks like this are also pretty expensive. So the fact that this is being called in the Update() of Blocks is not so good. I would suggest storing this in a variable on Awake() or caching it some other way.

AlfredoCuevas:
I noticed that you changed some of the Assets/Plugins ... meta files. I'm pretty sure these should not be changed. They relate to me using the Rider Editor instead of Visual Studio. Also I noticed that someone else moved the Editor folder into `Assets/Materials/Character Materials/Team Skins/`. I think this is the reason that Assets/Plugins/ is always being changed when it shouldn't be. After we have merged all the pull requests. I will make a Pull Request delete the meta files and fix this issue.

Assets/Scripts/Lobby/Player Sign In/ActivePlayers.cs
```diff
@@ -60,16 +60,35 @@ public static void UpdateScore(int playerIndex, int points)
         }
     }
 
+    public static void UpdateWins(int playerIndex)
+    {
+        int winScoreToUpdate = _activePlayers.First(p => p.PlayerNumber == playerIndex).Wins++;
+        if(winScoreToUpdate < 0)
+        {
+            winScoreToUpdate = 0;
+        }
```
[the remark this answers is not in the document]
JamesCoulter: I wasn't actually thinking about this at all, since this is how it was before I did anything. When I came into the script to add the new point functions, I just copied what was already there, and changed the function and variable names. I just assumed it worked fine since it never seemed to cause problems in the past. Looking at it now, both functions could just be one line then, right?

Assets/Scripts/UI/Match/RespawnUI.cs
JamesCoulter: Yeah, that makes sense

Assets/Scripts/UI/Match/UITreasureScore.cs
JamesCoulter: Oh, yeah, this script is not in use. I had made it for that first prototype awhile ago, but since then I got ride of the object to which it was attached. I'll delete it.

Assets/Scripts/Lobby/Player Sign In/ActivePlayers.cs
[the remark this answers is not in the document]
AlfredoCuevas: I think so yes. If all we want to do is update wins then you could just take the right half of line 65. However, if we did want to do an error check then we would probably want to save the player variable and then do the if statement. Ex)
 `var player = _activePlayer.First(p...);`
`player.Wins++;`
`if(player.Wins < 0) player.Wins =0;`

JamesCoulter:
I don't know if the conditional is actually needed, but I'll trust the judgment of whomever wrote the script in the first place and put that part back in like you said

PR #297, opened 2019-10-10
Treasure grab ui

JamesCoulter:
Contributor: Coulter

Goal: Add UI above player heads to track current number of points over total points needed to win the round.

Files Changed:
- Player.prefab: Added score text to the player canvas that appears above the player's head and tracks the number of points they have at any given time.
- TreasureUI.cs: Small script that changes the above UI text as appropriate.
- TreasureGrabModeManager.cs: Made into a singleton so that TreasureUI.cs could access the PointsRequired property.
- Treasure Grab Test.unity: Changed required win and point values to what is currently expected to be the default for the game mode.

How to test:
- Run the Treasure Grab Test.unity scene.
- Play through the game (break treasure blocks, kill players, etc.) to make sure that the score values above player heads are correct.
- Change required win and point values in [Treasure Grab Match Manager].prefab and then play again to ensure all UI (text above player heads and UI at end of match) still shows correct values.
- Make sure no UI weirdness was introduced into the other modes. Shouldn't be an issue, but still safe to check.

Notes: The exact positioning and size of the new UI can be changed pretty easily. Making it more like it was in the prototype build, where the new UI for all players was always at the top of the screen, would be harder, especially since even in the prototype there were bugs with that UI. I think we should stick with UI over player heads unless playtest feedback says to do otherwise.


Assets/Scripts/UI/Match/TreasureUI.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        ChangeText();
+    }
+
+    private void ChangeText()
+    {
+        _playerScore = ActivePlayers.GetScore((int)gameObject.GetComponentInParent<PlayerInput>().Index);
+        _scoreText.text = _playerScore.ToString() + "/" + TreasureGrabModeManager.Instance.PointsRequired.ToString();
+    }
+}
```
AlfredoCuevas: I'm going to have to veto this for the same reasons as the last pull request. It feels like there should be some kind of event for OnCoinBlockBroken() or OnCoinsCollected() or OnCoinChange(). GetComponentInParents() can be expensive and should not be called in Update(). Also, even without the GetComponent() this just isn't something that I think should be running every frame.

AlfredoCuevas: Actually, upon further inspection, it looks like these events already exist. `TreasureBlockBreaker.OnTreasureBlockBreak` and `SquishDetector.OnPlayerSquished`. I would recommend making a new event in TreasureGrabModeManager.cs called something like PlayerScoreUpdated. This event would be called in `TreasureGrabModeManager.UpdateScore`. This way you can be sure that the text is only updated after the score's are updated. You can even use the event to pass in `PointsRequired` into `ChangeText()`. That way you don't have to make it a Singleton (maybe in the future it will have to become a singleton but for now I would prefer finding ways around that)

Assets/Scripts/Singletons/GameManager.cs
```diff
@@ -14,10 +14,10 @@ public class GameManager : MonoBehaviour
     [SerializeField]
     private SceneTransitioner _sceneTransitioner;
 
-    public static GameManager Instance { get; private set; }
-
     public int WinsRequired { get { return _winsRequired; } }
 
+    public static GameManager Instance { get; private set; }
```
AlfredoCuevas: Did this only move the position of this line of code?

Assets/Scripts/Game Mode Managers/TreasureGrabModeManager.cs
```diff
@@ -14,14 +14,26 @@ public class TreasureGrabModeManager : MonoBehaviour
     private MatchEndManager _matchEndManager;
     private AlivePlayerTracker _alivePlayerTracker;
 
+    public int PointsRequired { get { return _pointsRequired; } }
+
+    public static TreasureGrabModeManager Instance { get; private set; }
```
AlfredoCuevas: We might be able to avoid making this a Singleton for now. Look at my other comment for details.

Assets/Scripts/Singletons/GameManager.cs
JamesCoulter: Yeah, I saw in other scripts we were always putting the Instance line last

Assets/Scripts/UI/Match/TreasureUI.cs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    private void OnDestroy()
+    {
+        TreasureGrabModeManager.OnPlayerScoreUpdated -= OnPlayerScoreUpdated_ChangeScoreText;
+    }
+
+    private void OnPlayerScoreUpdated_ChangeScoreText(int pointsRequired)
+    {
+        _playerScore = ActivePlayers.GetScore((int)gameObject.GetComponentInParent<PlayerInput>().Index);
```
AlfredoCuevas: I think that we could still save this GetComponent() call in a variable but other than that this looks good.

Assets/Scripts/Game Mode Managers/TreasureGrabModeManager.cs
```diff
@@ -16,12 +18,12 @@ public class TreasureGrabModeManager : MonoBehaviour
 
     private void Awake()
     {
+        ActivePlayers.ResetScores();
```
AlfredoCuevas: Why did you move this? I'm not questioning it just trying to understand the code better.

AlfredoCuevas:
Hey, I actually noticed this UI bug right now as I was doing a second test @JamesCoulter . 
Bug

It seems that sometimes the Respawn Icon doesn't disappear even after a player respawns. This doesn't happen every time. Only sometimes. Looks like it happens more often when respawn blocks break really quickly one after another

Assets/Scripts/Game Mode Managers/TreasureGrabModeManager.cs
JamesCoulter: I moved it because where it was before, the text would reset as soon as a player won the round. It didn't look right to me. So moving it here ensured that the scores at the end of the match would remain until the next match.

AlfredoCuevas:
Bug
I had a lot harder time replicating the bug this time around. However, this time. The ui stopped working AND the players never re-spawned.
* The way I replicated the bug was by getting all the players to die at the same time. The respawn blocks spawned stacked one after another. Then I just waited for a while. Maybe after 3 minutes of repeated dying it finally happened. I think it is VERY unlikely to happen in actual game play.

AlfredoCuevas:
Bug2
This bug looks like it might  be more complex that I initially thought. It happened again. Same as before, I really had to try to make it happen. However, I just noticed, that the not only did the player block not respawn. All blocks were no longer spawning.

JamesCoulter:
@AlfredoCuevas I actually haven't been able to reproduce the two most recent bugs you mention here (players not respawning and blocks no longer spawning). I also haven't been able to get the UI to mess up anymore. However, I am now experiencing the issue where two of the same player respawn much more frequently.

> 
> 
> Bug2
> This bug looks like it might be more complex that I initially thought. It happened again. Same as before, I really had to try to make it happen. However, I just noticed, that the not only did the player block not respawn. All blocks were no longer spawning.

Could this particular issue have been due to the simple fact that those towers of blocks reached the spawn zone? Because as long as blocks are in that zone, nothing will spawn

AlfredoCuevas:
> @AlfredoCuevas I actually haven't been able to reproduce the two most recent bugs you mention here (players not respawning and blocks no longer spawning). I also haven't been able to get the UI to mess up anymore. However, I am now experiencing the issue where two of the same player respawn much more frequently.
> 
> > Bug2
> > This bug looks like it might be more complex that I initially thought. It happened again. Same as before, I really had to try to make it happen. However, I just noticed, that the not only did the player block not respawn. All blocks were no longer spawning.
> 
> Could this particular issue have been due to the simple fact that those towers of blocks reached the spawn zone? Because as long as blocks are in that zone, nothing will spawn

The blocks no longer spawning makes sense. I forgot about the blocks getting too high. That is probably why they stopped.

PR #299, opened 2019-10-17
Update-Credits

AlfredoCuevas:
Contributors: Alfredo, Emma

Goal: Update the credits screen based on Emma's new design

Changes: 
* Lobby.unity; re-positioned the transforms

How to test: 
* Go into the credits scene in the lobby. Make sure it looks good on your screen.


AlfredoCuevas:
> I don't really know how credits should look for a project like this, and honestly I think I personally prefer how we had it before. But as far as whether or not the PR breaks the game, it works, so I'll approve it.

You can see the mock up on the gauntlet if you scroll up  enough. I'm just going off of that.

PR #304, opened 2019-10-25
Block Dunk Respawning

IanGRap:
Contributors: Ian

Goal: Add a new respawn system that didn't revolve around having to squish players as a block to rejoin the game. This also included a bit of respawning refactor.

Important Files:
- Respawner.cs: Handles functionality for respawning in matches.
- [BlockSpawner].prefab: Added in new respawner component.
- [TreasureGrabBlockSpawner].prefab: Added in new respawner component and set respawning to respawn as players.
-  Teams.cs: Added functions to get team color and designation from a player index.
- BlockSpawner.cs: Removed respawn functionality to be moved to respawner component.
- RespawnQueue.cs: Made it so you have to take the entire queue from the component when accessing it, instead of being able to change it from within another class.

- PlayerRespawner.cs: Added option for easy respawn on respawn blocks (land and you respawn).
- RespawnLight.cs: A light that spawns when a player respawn without having to use the respawn block. Helps to indicate when the player is.
- Other scripts: Set a bunch of code to use team color instead of skin color when in a team game mode.

How to test: Boot up block dunk and die a few times. Players should respawn as players with a light overhead. Players should spawn near their goal. Load up Treasure grab. When you die, you should spawn as a player near the center of the arena.

Notes: I had to change the treasure grab spawner or else it would break without the new respawner component. Let me know if I should roll back any changes in treasure grab.


Assets/Scripts/Blocks/Respawn/PlayerRespawner.cs
```diff
@@ -16,10 +16,15 @@ public class PlayerRespawner : BlockDamager
     [SerializeField]
     private GameObject _playerRespawnBeamPrefab;
 
+    [SerializeField]
+    private bool _easyRespawn = false;
```
AlfredoCuevas: Tool Tip?

Assets/Scripts/Blocks/Respawn/PlayerRespawner.cs
```diff
@@ -36,6 +41,11 @@ private void OnDestroy()
         MatchEndManager.OnMatchEnded -= OnMatchEnded_Break;
     }
 
+    public static void RespawnPlayer(GameObject player)
+    {
+        OnPlayerRespawn?.Invoke(player);
+    }
+
```
AlfredoCuevas: I'm still not sure what the ? does here. This is just personal question.

Assets/Scripts/Blocks/Respawn/PlayerRespawner.cs
```diff
@@ -16,10 +16,15 @@ public class PlayerRespawner : BlockDamager
...
 
+    [SerializeField]
+    private bool _easyRespawn = false;
+
     private Skin _playerSkin;
     private bool _playerSkinAssigned = false;
     private bool _isBreaking = false;
 
+    public bool HasKilledPlayer { get; set; }
+
```
AlfredoCuevas: Looks like both this and BlockDamager have a HasKilledPlayer variable. However, this one is only ever called from this script. so it never becomes true.

Assets/Scripts/Blocks/Respawn/PlayerRespawner.cs
AlfredoCuevas: easyRespawn is not self explanatory if you are just someone looking at it from the inspector.

Assets/Scripts/BlockSpawner/Respawner.cs
```diff
@@ -0,0 +1,138 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System.Linq;
+
+[RequireComponent(typeof(RespawnQueue))]
+public class Respawner : MonoBehaviour
```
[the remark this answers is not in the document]
JamesCoulter: This is a good point. It should be at least a little more descriptive

Assets/Scripts/Blocks/Respawn/PlayerRespawner.cs
JamesCoulter: Yeah, this line isn't necessary. It only hides the property in BlockDamager, which is the one that gets set to true in SquishDetector when a player is squished.

Assets/Scripts/BlockSpawner/BlockSpawner.cs
```diff
@@ -119,32 +117,7 @@ public void OnCoundownComplete_Spawn(bool countdownComplete)
...
-            respawnBlock.GetComponent<PlayerRespawner>().SetPlayerSkin(_playerToRespawn.Skin);
-
-            AudioManager.Instance.PlayOneShot("Block Spawn Rock");
-            _playerToRespawn.respawnImage.AssignTransform(respawnBlock.transform);
-            _respawnQueue.RemoveFromRespawnQueue(_playerToRespawn.Index);
-            yOffset += 1f;
-        }
-    }
-
+
```
JamesCoulter: I like moving the respawn stuff into its own script. I should have done this back when I was refactoring the spawner at the beginning of the summer

Assets/Scripts/UI/Match/RespawnImage.cs
```diff
@@ -77,7 +77,14 @@ public void AssignImage(Sprite sprite, Color color)
...
 
-        _playerIcon.color = new Color(_playerIcon.color.r, _playerIcon.color.g, _playerIcon.color.b, 0f);
+        if (GameManager.Instance.TeamMode)
+        {
+            _playerIcon.color = new Color(color.r, color.g, color.b, 0f);
+        }
+        else
+        {
+            _playerIcon.color = new Color(_playerIcon.color.r, _playerIcon.color.g, _playerIcon.color.b, 0f);
+        }
```
JamesCoulter: I'm not sure if I understand the need for this. Is it because the team skins are just normal colors, not special skins?

JamesCoulter:
I would mostly repeat Alfredo's requested changes. The HasKilledPlayer thing is the only issue that really needs to be fixed. That bug Alfredo found is also important, but I was unable to reproduce it.

Assets/Scripts/UI/Match/RespawnImage.cs
IanGRap: yeah basically. But I feel like it will need to change in the future

Assets/Scripts/Blocks/Core/BlockAdjacencyQueries.cs
```diff
@@ -59,7 +59,7 @@ public bool AdjacentBlockIsPushed(Vector3 direction)
     public bool FallingPathBlocked()
     {
         bool pathIsBlocked = 
-            GetAdjacentObjects(Vector3.down).FirstOrDefault(o => 
+            GetAdjacentObjects(Vector3.down).Where(o => o != null).FirstOrDefault(o => 
             o.tag == "Environment" ||
```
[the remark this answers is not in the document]
IanGRap: I've been noticing that for a bit but I don't think it is related to this stuff.

Assets/Scripts/Blocks/Respawn/PlayerRespawner.cs
IanGRap: It is supposed to check if the value is null before calling the function I believe.

AlfredoCuevas:
I looked it over and the spawning bug seems to be gone.

PR #305, opened 2019-10-29
Treasure Grab Coin Visual

JamesCoulter:
Contributors: Coulter, Athene

Goal: Add a coin that flies out of broken treasure chests and into the players that break them.

Important Files:
- Treasure Coin.prefab: Has a particle effect and a coin model.
- TreasureCoinMovement.cs: Script to make the coin move in such a way that it doesn't just clip through blocks and fly straight to the player. I did a lot of experimenting with DOTween functions and lerping, and the final method I reached that seemed to look the best involves the coin first moving out toward the screen with DOMove(), and then once that finishes I use Lerp to make sure the coin follows the player over time before it disappears, rather than just going to one of the player's previous positions.
- TreasureCoinRotation.cs: Makes the coin's model rotate over time. For whatever reason, a high rotation speed multiplier is needed to make the coin rotate at a noticeable rate.
- PlayerDash.cs: Spawns the coin when a chest is broken.
- Treasure Block and Coin particles: Slightly changed to hopefully look a bit better. Will change back if others disagree. The block particle in particular was changed to hopefully look more like a glimmering aura, instead of the upward-flying particles that used to be there.

How to test:
- Start up the Treasure Grab scene.
- Break chests.
- Observe the coin that spawns. Is the whole visual aspect of it satisfying? Does its movement look weird and/or janky, or do you think it works?
- Observe the treasure block particle effect. Do you like it more than before? Less? Are you Indifferent?

Notes: I was originally trying to use DOPath for the entire movement path of the coin because that function would have been perfect, but it seems to only be called a single time, meaning that the coin will only go toward the point it was given when the function called, even if the player's current position is different. This was the reason for me deciding that the second half of the movement should be a Lerp called in Update(). But I still liked having the coin first move away from the arena so that it wouldn't phase through blocks, and so that it would be super noticeable for players. If anyone has better ideas about how the coin could be made immediately clear to players instead, I would be willing to try it.
Also, the coin model will receive at least one more iteration, probably around a week from now. I figured it would be fine adding that in separately later on so that I could get other stuff done in the meantime.


Assets/Scripts/Blocks/Treasure/TreasureCoinMovement.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    // Makes the coin jump out to its halfway point during the movement from chest to player.
+    public void MoveCoinFirstHalf()
+    {
+        Vector3 midPoint = CalculateMidPoint(transform.position, gameObject.transform.parent.position);
+        transform.DOMove(midPoint, _firstHalfOfMovementDuration);
+        _readyToLerp = true;
+    }
```
AlfredoCuevas: I'm not sure if this is what you were intending but I think you probably want 
`transform.DOMove(midPoint, _firstHalfOfMovementDuration).OnComplete(() =>
        {
            _readyToLerp = true;
        });` 
This makes sure to wait until the DoMove is done before making _readyToLerp = True. That way the DoMove and the Vector3.Lerp in update aren't fighting each other.

JamesCoulter:
@AlfredoCuevas I think at this point my plan is to simply have all special sound effects related to the coins just be included altogether in a later PR. I did address your other points, though.

PR #307, opened 2019-10-30
Block Dunk Points for Kill

IanGRap:
Contributors: Ian

Goal: Add the functionality to let players score points in block dunk for squishing players on the other team.

Important Files:
- PointMarker.cs: A floating number that appears when players are squished to indicate how many points their death was worth.
- BlockDunkModeManager.cs: Adds points to when players are squished.

How to test: Play some block dunk. When people get squished, a number should appear where they died. The opposing team should get that many points. If you get enough points with squishing other players, the game should end.


Assets/PointMarker.cs
```diff
@@ -0,0 +1,22 @@
...
+
+public class PointMarker : MonoBehaviour
+{
+    public void SetMarker(Color color)
+    {
+        GetComponentInChildren<TextMeshPro>().color = color;
+
+        transform.DOMove(transform.position + (Vector3.up * 1.1f), 1f);
+
+        Invoke("DestroySelf", 1f);
```
AlfredoCuevas: I'm pretty sure that you can do `transform.DOMove(transform.position + (Vector3.up * 1.1f), 1f).OnComplete(() => { DestroySelf(); });`
* Note: OnComplete is not the same as onComplete. I made that mistake 1 time too many.

This way you don't have to do an Invoke() call. The destroy will be called at the end of the DoMove() effect.

Assets/Scripts/Game Mode Managers/BlockDunkModeManager.cs
```diff
@@ -56,6 +62,19 @@ public void AddPointsToScore(TeamDesignations team, int value)
...
+        {
+            SquishDetector.OnPlayerSquished += PointsForSquish;
+        }
+    }
+
+    private void OnDestroy()
+    {
+        SquishDetector.OnPlayerSquished -= PointsForSquish;
+    }
+
```
AlfredoCuevas: Should OnDestroy() also have a `if(_pointsForSquish)` check

PR #310, opened 2019-11-14
Background Eyes Test

JamesCoulter:
Contributors: Coulter, Athene

Goal: Create an effect where glowing eyes can be seen opening and closing in the background of the new arena a la old scooby-doo-era cartoons.

Important Files:
- Background Eyes 1.prefab: Holds eye sprite, plus two masks that go up and down to cover it. Also has a 2D collider which is used to detect when a new potential spawn position overlaps with another eye that has already spawned.
- EyeScript.cs: Sets the eye sprite randomly from a list, controls the movement of the masks. Several variables are available to tweak so that the masks act as much like eyelashes as possible. If other differently-shaped eye sprites are introduced, they could be put in separate prefabs where the sprite mask movement values can be tweaked to fit the eye shape.
- Eye Zone.prefab: Empty game object that spawns eyes using randomized positions and spawn time intervals.
- EyeSpawner.cs: Script that controls when and where to spawn eyes.

How to test:
- Either pick a game mode scene or make a new empty one.
- Drag the Eye Zone prefab into the scene. If you put the eyes in an existing scene, it could be a good way to judge it while playing. Also, if you put the zone in an existing scene, make sure to put it a little behind everything on the z axis so that it doesn't overlap with blocks and players.
- Play the scene.
- Observe the eyes spawning, opening, closing, and being destroyed. Ideally, no eyes should be overlapping each other. Do you like how often they spawn, how quickly they open and close, etc?


Assets/Scripts/Environment/Visual/EyeSpawner.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    // Set a random spawn position. If the chosen position is too close to another eye that has already been spawned, run again
+    // to pick a new position. This ensures that eyes and their masks never end up overlapping each other.
+    private void FindSpawnPosition()
+    {
+        _xSpawnPos = Random.Range(gameObject.transform.position.x - _eyeSpawnRange, gameObject.transform.position.x + _eyeSpawnRange);
+        _ySpawnPos = Random.Range(gameObject.transform.position.y - _eyeSpawnRange, gameObject.transform.position.y + _eyeSpawnRange);
+        _spawnPosition = new Vector3(_xSpawnPos, _ySpawnPos, 0f);
+        _spawnCollisionRange = new Vector2(_xSpawnCollisionRange, _ySpawnCollisionRange);
```
AlfredoCuevas: Do _xSpawnCollisionRange and _ySpawnCollisionRange ever change? If not then I think you can set _spawnCollisionRange inside of Start() or Awake() and not reset it.

Assets/Scripts/Environment/Visual/EyeSpawner.cs
```diff
@@ -0,0 +1,63 @@
...
+        if (Physics2D.OverlapBox(_spawnPosition, _spawnCollisionRange, 0) != null)
+        {
+            FindSpawnPosition();
+        }
```
AlfredoCuevas: Does this if statement also include non eye ball stuff (like blocks) that have colliders? 

This might be small but I think that a do..while() loop might be better here. If I remember correctly, Calling a function recursively causes some memory usage. However, if you do a while loop, execution will simply jump back to the top of the function, no extra memory used. But, since this likely called <2 times. It probably doesn't matter.

Assets/Scripts/Environment/Visual/EyeSpawner.cs
```diff
@@ -5,59 +5,78 @@
...
+        {
+            FindSpawnPosition();
+            for (int i = 0; i < _eyes.Count; i++)
+            {
+                if (!_eyes[i].activeSelf)
+                {
+                    _eyes[i].SetActive(true);
+                    _eyes[i].transform.position = _spawnPosition;
+                    _eyes[i].transform.SetParent(gameObject.transform);
+                    _eyes[i].transform.localScale = _eyes[i].transform.localScale * Random.Range(_minScale, _maxScale);
```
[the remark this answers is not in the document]
JamesCoulter: These comments actually made me realize, the eyes in the list are already childed to the zone prefab, so that `SetParent()` line is unnecessary. You also made me realize that I was constantly multiplying the eye's scale over time, meaning that after awhile all the eyes would become too small to see. So I also fixed that.

PR #311, opened 2019-11-19
Lightweight Rendering Pipeline + Material Fixes

SolSearcher:
Goal : Upgrade the project to the lightweight rendering pipeline so we can use the color changes with baked lighting as well as have awesome effects.

What I changed : 

- Added a lightweight render pipeline asset in the rendering folder and set the project up to use it
- Converted all materials to lightweight render pipeline compatible materials 
- Redid the landscape in the lobby scene since the the terrain got messed up with the new pipeline
- Went through many scenes to see if any materials weren't fixed. Thus the scene changes

How to Test:
- Load up the branch and try to play the game. 
- Look out for bright pink textures, that means the material didn't get converted. This is usually because the material being used was embedded with the model so it just needs to be turned into a Unity material.
- Look out for any other visual glitches


JamesCoulter:
Okay, so I was also getting un-converted materials, apparently the same ones as Alfredo. Several of the hats in the lobby were pink, lack of effects from the bomb blocks. The roof object of all the arenas also had an undefined material. I think everything else looked okay. I didn't notice anything else, at least.

PR #312, opened 2019-12-17
New dash sounds

AlfredoCuevas:
Contributors: Fernando <sub> Alfredo </sub>

What changed: 
* New dash sound files: the old Character Dash audio clip was deleted and 4 new ones were added
* DashRandomizer.cs: This new script lives on the Character Dash gameobject that holds the AudioSource component for the dash sound. It is not directly on the AudioManager. It has one public function that will play the Dash clip and then swap it out for another dash clip
* AudioManager.cs: Has a reference to the DashRandomizer and uses it to call a new PlayDashSound() function
* PlayerDash.cs: calls the new dash function instead of the old PlayOneShot("") function
* AudioManager.prefab: updated to hold all the new audio clips and scripts. The scene was not changed only the prefab to prevent any conflicts

How to test: 
* start in any scene and dash around
* The dash sounds should randomly pick one of 4 different clips
* How does it sound? It should be a subtle difference. The dash shouldn't really be grabbing the players attention too much. 
* Ensure that there are no errors that pop up in the console when dashing around.


Assets/Scripts/Singletons/AudioManager.cs
```diff
@@ -93,6 +96,11 @@ public void Stop(string soundGameObjectName)
         sound.Stop();
     }
 
+    public void PlayDashSound()
+    {
+        _dashRandomizer.PlayDashSound();
+    }
+
```
JamesCoulter: It seems weird to me that a script calls another script that calls another script just to run a single simple function, but at the same time I'm not sure at this point what the better way would be

PR #319, opened 2020-01-22
Lobby Visual Update and Fog Update

SolSearcher:
Contributors: Reed

Changes:

- LobbyStateManager.cs - Just added in some fog scripts to cleanly fade in / out the fog

- Lobby Scene - Lots of Environment assets came in, added fog into the Camera Manager (not the whole prefab just this instance) 

- Player Prefab - made the player light not effect players since it was too bright

- ColorSelector.cs - Made it work for fog

- Match Scene and Block Dunk Scene - Added color change to the fog

How to Test:

- Go through the lobby and make sure no spots are way too bright. This would be an intensity value being too high for the HDR

- Make sure no error messages pop up

- Most of the changes were just the environment gameobjects and I used a prefab to bring things over


Assets/Scripts/Lobby/State Management/LobbyStateManager.cs
```diff
@@ -202,4 +205,22 @@ private void KeepOnlyOneInstance()
             return;
         }
     }
+
+    private void FogEnable(){
+        if(_camera.transform.Find("Fog Camera")){
+            // Used for fading in / out Main Menu and character select fog
+            ParticleSystem fogParticleSystem = _camera.transform.Find("Fog Camera").GetComponent<ParticleSystem>();
```
JamesCoulter: A small optimization here could be to make a variable and use it like:
`Transform fog = _camera.transform.Find("Fog Camera");`
`if(fog)`
`{`
`ParticleSystem fogParticleSystem = fog.GetComponent<ParticleSystem>();`
`}`
This way you'll only call Find() once, which is good because Find() is fairly performance-heavy. And in general we try to avoid using it. I think it's fine in this case just because the call is very infrequent, but we should try to minimize its usage as much as possible

JamesCoulter: Also, a quick style thing: we've been following the practice of having all curly braces on their own lines, so you should fix that in these functions really quick for the sake of consistency.

SolSearcher: Addressed your fixes. Made everything in line with the style guide. I made a serialized field for the two fog volumes that are being turned on / off. This clutters up the Lobby state manager more than the one attached to the camera did but at least it doesn't need "Find()" operations and the fog isn't attached to the camera anymore.

Assets/Scripts/Lobby/State Management/LobbyStateManager.cs
```diff
@@ -92,6 +98,7 @@ public void OpenCharacterSelectMenuTeams()
         _destinationRotation = _characterMenuCameraRotation;
         GameState.SetLobbyState(LobbyState.CharacterMenuTeams);
         _lerpSpeed = 0.14f;
+        FogEnable(_fogCharacterSelect);
```
JamesCoulter: I noticed that FogEnable() isn't doing anything because _fogCharacterSelect is already going by default once the scene starts. The simplest fix would either be to add a FodDisable() call inside Awake(), or just get rid of the FogEnable() function, since it is currently redundant.

SolSearcher: It actually does something. The reason I added the fog onto the camera in the first place is that when you were in the tutorial scene you could still see the character select fog on the top of the screen which looked really awkward. Unfortunately you can still see this effect for a tiny bit when you go from character select to the tutorial but FogDisable() fades out the fog cleanly allowing for the intended look of the tutorial. 

To test this go into the Lobby scene, hit play, choose any gamemode, add two or more players and then instead of hitting play go into the tutorial, the fog will disappear, now take both characters to the left to go back to the main menu stage. Repeat before steps and this time when you come in from the menu to the character select FogEnable() will trigger and re-enable the character select fog.

JamesCoulter: Ooooohhhhh I see now, thank you for clearing that up. I wasn't considering the case of people leaving the tutorial and then going back to the character screen again.

SolSearcher: Gotcha! Thanks for saying something about the camera fog, this looks so much better

